PR #9, opened 2016-05-27
Add `lift` to `FutureEffect`

pavlosgi:


shared/src/main/scala/org/atnos/eff/FutureEffect.scala
```diff
@@ -24,6 +24,15 @@ trait FutureCreation {
 
   def async[R, A](a: =>A)(implicit m: Member[Future, R], ec: ExecutionContext): Eff[R, A] =
     send(Future(a))
+
+  def lift[R, A](
```
etorreborre: We'd better call this `liftFuture` to avoid having import conflicts if we adopt a similar technique with another effect and want to import `org.atnos.eff.all._`.

etorreborre: Also in terms of coding style I generally put everything on the same line unless it goes above 120 chars. And I don't use accolades for 1-line statements.

etorreborre: Finally it would be nice to have a bit of syntax in `syntax.future` to be able to write `myFuture.liftFuture`.

etorreborre: Can you also use `liftFuture` in the `FutureEffectSpec` so that it is tested?

etorreborre:
Brilliant, thanks!

etorreborre:
Depend on `1.7-20160526110310-1704d42` if you want to use your nice new toy :-).

pavlosgi:
👍

PR #10, opened 2016-05-27
Modify FiniteDuration to Duration in FutureEffect

pavlosgi:
Can we add this ?


etorreborre:
Definitely.

etorreborre:
In `1.7-20160527140737-097745b`.

PR #12, opened 2016-07-11
Minor doc fix

benhutchison:
(Note there's also a strange line at end of page that doesn't make sense to me:
Then we can use this effect in a computation:`> 5`)


etorreborre:
Thanks, those are generated pages. On master I added your typo to the class generating this page but also I fixed the example at the end of the page.

PR #23, opened 2016-11-03
Add eval.defer and task.suspend for trampolining

edmundnoble:
Also found that the other test in `EvalEffectSpec` wasn't being executed.


edmundnoble:
I've since also added stack-safety for `monix.Task` to this PR.

etorreborre:
Thanks for this PR it's great!

edmundnoble:
@etorreborre you deleted `TaskEffect` but didn't add this functionality into `Async`. Is there a chance it could be added back in, perhaps somewhere else if `Async` does not need it? I know `scalaz.concurrent.Task` and `fs2.Task` could use it as well.

etorreborre:
That's right I forgot to add it (I'm doing too much at the same time), I'll do it later on today. Thanks for watching my back.

PR #24, opened 2016-11-06
Make ops classes final and extend AnyVal

edmundnoble:
I am not sure how much this will matter in dotty, but this should minimize allocations and be more friendly to the inliner.

### DIFF
--- a/monix/shared/src/main/scala/org/atnos/eff/monix/AsyncTaskService.scala
+++ b/monix/shared/src/main/scala/org/atnos/eff/monix/AsyncTaskService.scala
@@ -100,7 +100,7 @@ object AsyncTaskService extends AsyncTaskServiceInterpretation {
 
 }
 
-case class AsyncTask[A](task: Task[A]) extends Async[A] {
+case class AsyncTask[A](task: Task[A]) extends AnyVal with Async[A] {
   def attempt: Async[Throwable Either A] =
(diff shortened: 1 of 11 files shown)


etorreborre:
The compilation failed for the `monix` and `scalaz` modules, can you please have a look at that? Otherwise I think this is a good idea :-).

edmundnoble:
I believe there's some intermittent failures in monix and scalaz task tests affecting this build.

etorreborre:
Can you rebase with master because I did some changes today to stabilize the build? Thanks.

edmundnoble:
Rebased.

monix/shared/src/main/scala/org/atnos/eff/monix/AsyncTaskService.scala
```diff
@@ -40,7 +40,7 @@ case class AsyncTaskService() extends AsyncService {
 trait AsyncTaskServiceInterpretation {
 
   def runAsyncTask[A](e: Eff[Fx.fx1[Async], A]): Task[A] =
-    e.detachA(ApplicativeAsync) match {
```
edmundnoble: Ohhhh my, I see. The monad is getting picked by implicit resolution. Gosh darn it :P

etorreborre:
Thanks!

PR #25, opened 2016-11-06
Add >> and << to Eff so they can be used applicatively

edmundnoble:
Similarly to `*>` and `<*`.


etorreborre:
I would like to keep `*>` for the applicative case and `>>` for the sequential one because I think they both have their uses. I redefined `*>` directly on `Eff` because it was otherwise implemented in terms of `flatMap` and was forcing sequential evaluation.

edmundnoble:
This makes sense to me.

PR #27, opened 2016-11-11
Run callback on completion instead of on cancel

edmundnoble:
`AsyncTaskService.async` runs the callback on task cancel, not on task completion. This is a fix for that.


edmundnoble:
Obsolete. Nice work on the new async!

PR #28, opened 2016-11-12
Add flatTraverseA and flatSequenceA

edmundnoble:
Adds the `flatTraverseA` and `flatSequenceA` operators based on the `flatTraverse` and `flatSequence` operators in cats.


etorreborre:
Thanks :-)

PR #30, opened 2016-11-28
eff-cats => eff name migration

bepcyc:
According to documentation and artifacts published yesterday (27.11.2016) it's no longer `eff-cats` project but `eff` project.
Renaming makes this page unusable because of broken URLs and artifact names/versions. I fixed that as much as possible to a working state as I'm using this page sometimes.
Feel free to reject the PR as I was just trying to make this page usable again for myself.


etorreborre:
I'm sorry I just forgot to push...

PR #32, opened 2016-12-06
Fix Arrs.mapLast for empty case, remove unapplySeq calls

edmundnoble:
`Arrs.mapLast` doesn't actually apply the function it's passed if the `Arrs` is empty. This fixes that, removes a couple of `unapplySeq` calls in `go` in favor of explicit `isEmpty` and pulls the `f(v)` call out of the branch. As well, replaces varargs `Vector(A*)` calls to avoid intermediate collections and replaces a call to `EffMonad[R].pure[A]` with `Eff.pure[R, A]` to make IntelliJ happy for now; it otherwise complains as it does not recognize kind projector in cross-projects (because it doesn't really support them). Let me know which changes are welcome. 

I also have some benchmarks pending for data structures for free monads if you are interested (especially implemented with reflection without remorse) at edmundnoble/scabsbench, which may be the basis for further PRs.

### DIFF
--- a/shared/src/main/scala/org/atnos/eff/Arrs.scala
+++ b/shared/src/main/scala/org/atnos/eff/Arrs.scala
@@ -26,7 +26,7 @@ case class Arrs[R, A, B](functions: Vector[Any => Eff[R, Any]]) extends (A => Ef
   /** map the last returned effect */
   def mapLast(f: Eff[R, B] => Eff[R, B]): Arrs[R, A, B] =
     functions match {
-      case Vector() => this
+      case v if v.isEmpty => Arrs[R, A, B](v :+ ((a: Any) => f(Eff.pure(a.asInstanceOf[B])).asInstanceOf[Eff[R, Any]]))
       case fs :+ last => Arrs(fs :+ ((x: Any) => f(last(x).asInstanceOf[Eff[R, B]]).asInstanceOf[Eff[R, Any]]))
     }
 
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/org/atnos/eff/Arrs.scala
```diff
@@ -26,7 +26,7 @@ case class Arrs[R, A, B](functions: Vector[Any => Eff[R, Any]]) extends (A => Ef
   /** map the last returned effect */
   def mapLast(f: Eff[R, B] => Eff[R, B]): Arrs[R, A, B] =
     functions match {
-      case Vector() => this
+      case v if v.isEmpty => Arrs[R, A, B](v :+ (a => f(Eff.pure(a)).asInstanceOf[Eff[R, Any]]))
```
etorreborre: Nice, could you please add a test for that fix?

etorreborre: Also you need to add a parameter type in there:
```
[error] /home/travis/build/atnos-org/eff/shared/src/main/scala/org/atnos/eff/Arrs.scala:29: missing parameter type
[error]       case v if v.isEmpty => Arrs[R, A, B](v :+ (a => f(Eff.pure(a)).asInstanceOf[Eff[R, Any]]))
[error] /home/travis/build/atnos-org/eff/shared/src/main/scala/org/atnos/eff/Arrs.scala:29: missing parameter type
[error]       case v if v.isEmpty => Arrs[R, A, B](v :+ (a => f(Eff.pure(a)).asInstanceOf[Eff[R, Any]]))
[error]                                                  ^ 
```

etorreborre:
Thanks, if you can add a test that's even better! Benchmarks are obviously very useful too. If we can improve some performance without making the code not too complex (it already is), that would be nice.

shared/src/main/scala/org/atnos/eff/Arrs.scala
edmundnoble: Absolutely!

edmundnoble: Damn, wish I'd added the test earlier now ;)

etorreborre:
Great thanks! Do you need a `2.0.2` soon?

edmundnoble:
💯 👍 :shipit:

etorreborre:
Done.

PR #35, opened 2016-12-15
fixed link

dyk:


etorreborre:
Can you also do the modification here please: https://github.com/atnos-org/eff/blob/master/jvm/src/test/scala/org/atnos/site/Tutorial.scala#L7?

dyk:
Done. New PR created

PR #42, opened 2016-12-23
remind readers to rename the smart constructor 'get'; name conflict i…

raygit:
…n the State monad. This name conflict is immediately evident when the reader copies-pastes (like i did) of the earlier tutorial on the impure interpreter; best to rename the smart constructor


etorreborre:
So I close this one which was integrated in the other PR.

PR #44, opened 2016-12-24
Tutorial changes

raygit:
I might be creating a pull request for a previous one; my bad.

When using `eff-cats` verison `2.2.0`, i discovered that `run` isn't necessary though the tutorial says it does.


etorreborre:
Thanks!

PR #47, opened 2016-12-27
allow to catch only specific exceptions

jozic:
couple changes here
- allow to catch and handle only specific exceptions (like it's possible with try/catch)
- changing SafeSpec to use only Safe effect - I was reading the spec to understand methods better and didn't get why option effect is needed there. Let me know if it was there for a purpose and i can resurrect it

### DIFF
--- a/jvm/src/test/scala/org/atnos/eff/SafeSpec.scala
+++ b/jvm/src/test/scala/org/atnos/eff/SafeSpec.scala
@@ -1,13 +1,13 @@
 package org.atnos.eff
 
-import cats._
-import data._
+import cats.Eval
 import cats.implicits._
 import org.atnos.eff.all._
 import org.atnos.eff.syntax.all._
 import org.specs2._
 import org.specs2.matcher.{Matcher, ThrownExpectations}
 import org.scalacheck.Gen
+
 import scala.collection.mutable.ListBuffer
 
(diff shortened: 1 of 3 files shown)


etorreborre:
Hi, thanks for the PR!

I think we should name `catchException` `recoverThrowable` instead since we are still dealing with `Throwables`. And that naming is similar with `recover` for futures for example.

Also I have added `Option` in the stack for `Safe` to provide an additional level of testing for the safe interpreters which are using different code paths when there are several effects at play.

jozic:
Hi Eric, thanks for the feedback!
I will rename `catchException` to `recoverThrowable` (or should it be just `recover`?)
Also you want Option back in SafeSpec, right? Another option can be to have separate a spec for Safe and another one for integrations with option and maybe some other simple effects. Your call.

etorreborre:
Maybe to make it less confusing you could leave things as they are now in your PR and add just one explicit integration test in `SafeSpec` to test that `catchException` doesn't fall over when mixed with other effects.

jozic:
sounds good, will update the PR soon.

jozic:
@etorreborre renamed as discussed and added new test mixing safe with option and eval

etorreborre:
Great, thanks!

PR #49, opened 2017-01-04
Fixed the memoization of async values.

### LINKED ISSUE #48: Async calls with continuations are not memoized correctly with a memo effect
Consider this example for `AsyncFutureInterpreterSpec`:
```scala
def e14 = {
    var invocationsNumber = 0
    val cache = ConcurrentHashMapCache()

    type S = Fx.fx2[Memoized, Async]
    val intAsync: Eff[S, Int] = asyncFork[S, String]("a").flatMap( x => asyncFork[S, Int]({ invocationsNumber += 1; 1 }))
    def makeRequest = asyncMemoized("only once", intAsync)

    (makeRequest >> makeRequest).runAsyncMemo(cache).runAsyncFuture must be_==(1).await
    invocationsNumber must be_==(1)
  }
```
Looks like `flatMap`ping causes the result of the first async to be cached. I would expect `1` to be cached, however `"a"` is cached instead:
```
[error]  x Async calls with continuations can be memoized with a memo effect
[error]  'a' is not equal to '1' (AsyncFutureInterpreterSpec.scala:179)

```

etorreborre:
closes #48.

It turns out that memoizing effects was a bit harder than I previously thought. Basically to memoize `Eff[R, A]` for a given effect `M` (`M |= R`) we need to cache all the `M[X]` values leading to the computation of `Eff[R, A]`.

This PR:

 - fixes the type parameter on `Cache` (as in PR #46)
 - introduces a new `SequenceCache` (and corresponding `MemoizedSeq` effect but I'm not sure it is that useful outside of this use case)
 - adds a new `memoizeEffect` interpreter to cache an effect `M` being used in an `Eff` computation

@edmundnoble thanks for looking at this PR. Most of the code is independent from `Async` and should be reusable in your PR.


etorreborre:
Please don't merge now. I need to add some reset functionality for a given key.

etorreborre:
In general I think it is better to leave the original committer merge once approved.

edmundnoble:
Sure, sounds good to me.

PR #50, opened 2017-01-07
Better support for discriminating between multiple Either types when interpreting

benhutchison:


edmundnoble:
Looks like the same test failure we saw on my PR. This looks great, but I'm afraid it's not source-compatible (only barely, I don't think many people specify the type param on runEither but I know I have to to prevent the compiler exploding). @etorreborre what's our position on source compatibility as yet?

etorreborre:
I think this is a useful evolution of the API and we should enforce the `U` pattern to the other effects with 2 type parameters like `State` and `Writer`.

@edmundnoble I think we should allow ourselves some flexibility on source compatibility at this stage. In that case I think the benefits outweigh the drawbacks and fixing broken code is fairly easy.

etorreborre:
👍

etorreborre:
@edmundnoble @benhutchison I think the failing tests just have a [`retries = 1` instead of `retries = 5`](https://github.com/atnos-org/eff/pull/46/files#diff-b6e7ef135074ad7cf2d9987cd498c7d5R68).

edmundnoble:
This gets a :+1: from me as well then.

etorreborre:
@benhutchison would you mind adding the corresponding `U` methods for `runState` and `runWriter` before I merge this?

etorreborre:
Thanks @benhutchison :1st_place_medal:

PR #51, opened 2017-01-08
Updated the try-eff script

etorreborre:
FYI @edmundnoble


jvm/src/test/scala/org/atnos/eff/AsyncFuturesSpec.scala
```diff
@@ -78,7 +78,7 @@ class AsyncFuturesSpec(implicit ee: ExecutionEnv) extends Specification with Sca
       }
 
     val run = Eff.traverseA(ls)(i => action[S](i))
-    eventually(retries = 1, sleep = 1.second) {
```
etorreborre: While I'm here...

PR #55, opened 2017-01-18
Disable tests in JavaScript by default

edmundnoble:
Until specs2 supports it. This consistently removes two seconds of overhead from `test` and `testQuick` (on my laptop), because starting three `node` instances is just that slow 😉


etorreborre:
👍 There's still a long way to go until specs2 supports JavaScript but hopefully one day...

PR #57, opened 2017-02-01
Add interceptNatM combinator

edmundnoble:
An effort to reconcile some copypasta. I still need to see if I can make this work with `memoizeEffectSequence` and friends.

Edit: Should I perhaps just call it `interceptM` or `flatIntercept`?

### DIFF
--- a/shared/src/main/scala/org/atnos/eff/Interpret.scala
+++ b/shared/src/main/scala/org/atnos/eff/Interpret.scala
@@ -581,6 +581,57 @@ trait Interpret {
     }
   }
 
+  type of[F[_], G[_]] = {type l[A] = F[G[A]]}
+
+  /**
+   * Intercept the values for one effect,
+   * emitting new values for the same effect inside a monad which is interleaved in
+   */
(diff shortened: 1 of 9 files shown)


shared/src/main/scala/org/atnos/eff/Interpret.scala
```diff
@@ -581,6 +581,57 @@ trait Interpret {
     }
   }
 
+  type of[F[_], G[_]] = {type l[A] = F[G[A]]}
+
+  /**
+   * Intercept the values for one effect and transform them into
+   * other values for the same effect
+   */
+  def interleaveMonadTransformation[R, T[_], F[_], A](effects: Eff[R, A], nat: T ~> (T `of` F)#l)
```
etorreborre: This looks like a net win for me. I would however create one additional method specializing `interleaveMonadTransformation` to the case where `F = Id` and use that simpler method where it can be used.

etorreborre:
Yes I would also prefer `interceptM` and simply `intercept` or `interceptId` for the case `F = Id`.

etorreborre:
Wrong button...

etorreborre:
👍

PR #58, opened 2017-02-02
Makes Unions use Vector instead of List for asymptotic reasons

edmundnoble:
`ap` has linear complexity, and `traverse` quadratic, if we use `List` in `Unions` and rely on `List` appending for `ap` concatenation. `Vector` can fix this; perhaps eventually `Vector` should be replaced with something `Catenable` for more performance. #


etorreborre:
Very welcome change!

PR #59, opened 2017-02-02
Indexed coproducts

edmundnoble:
Replaces `Union1`, `Union2L/R`, `Union3L/M/R` with a single representation `UnionTagged`, so that many anonymous implicits can be removed, some branches can be removed, and some `Member` and `IntoPoly` implicits can avoid allocating. I went with making indices increase from the right, starting at 1 on the right. So `Fx1[T] => Fx2[R, T]` is only a cast, for example.

### DIFF
--- a/shared/src/main/scala/org/atnos/eff/Eff.scala
+++ b/shared/src/main/scala/org/atnos/eff/Eff.scala
@@ -74,8 +74,8 @@ sealed trait Eff[R, A] {
   def flatMap[B](f: A => Eff[R, B]): Eff[R, B] =
     EffMonad[R].flatMap(this)(f)
 
-  def flatten[B](implicit ev: A =:= Eff[R, B]): Eff[R, B] =
-    flatMap(a => a)
+  def flatten[B](implicit ev: A <:< Eff[R, B]): Eff[R, B] =
+    flatMap(ev(_))
 
   /** add one last action to be executed after any computation chained to this Eff value */
   def addLast(l: =>Eff[R, Unit]): Eff[R, A] =
@@ -149,7 +149,7 @@ trait EffImplicits {
   final val effMonadUnsafe: Monad[Eff[AnyRef, ?]] = new Monad[Eff[AnyRef, ?]] {
 
     def pure[A](a: A): Eff[AnyRef, A] =
-      Pure(a)
+      Pure[AnyRef, A](a)
 
     override def map[A, B](fa: Eff[AnyRef, A])(f: A => B): Eff[AnyRef, B] =
       fa match {
@@ -190,7 +190,7 @@ trait EffImplicits {
   final val effApplicativeUnsafe: Applicative[Eff[AnyRef, ?]] = new Applicative[Eff[AnyRef, ?]] {
 
     def pure[A](a: A): Eff[AnyRef, A] =
-      Pure(a)
+      Pure[AnyRef, A](a)
 
     override def product[A, B](fa: Eff[AnyRef, A], fb: Eff[AnyRef, B]): Eff[AnyRef, (A, B)] =
       ap(map(fb)(b => (a: A) => (a, b)))(fa)
@@ -306,20 +306,18 @@ trait EffInterpretation {
       case Pure(a, Last(None))    => monad.pure(Right(a))
 
       case Impure(u, continuation, last) =>
-        u match {
-          case Union1(ta) =>
-            last match {
-              case Last(Some(l)) => ta.map(x => Left(continuation(x).addLast(last)))
-              case Last(None)    => ta.map(x => Left(continuation(x)))
-            }
+        val ta = u.tagged.valueUnsafe.asInstanceOf[M[A]]
+        last match {
+          case Last(Some(l)) => ta.map(x => Left(continuation(x).addLast(last)))
+          case Last(None)    => ta.map(x => Left(continuation(x)))
         }
 
       case ap @ ImpureAp(unions, continuation, last) =>
-        val effects = unions.unions.collect { case Union1(mx) => mx }
+        val effects = unions.unions.map(_.tagged.valueUnsafe.asInstanceOf[M[Any]])
         val sequenced = applicative.sequence(effects)
 
         last match {
-          case Last(Some(l)) => sequenced.map(x => Left(continuation(x).addLast(last)))
+          case Last(Some(_)) => sequenced.map(x => Left(continuation(x).addLast(last)))
           case Last(None)    => sequenced.map(x => Left(continuation(x)))
         }
     }
--- a/shared/src/main/scala/org/atnos/eff/IntoPoly.scala
+++ b/shared/src/main/scala/org/atnos/eff/IntoPoly.scala
@@ -26,7 +26,7 @@ trait IntoPolyLower1 extends IntoPolyLower2 {
     new IntoPoly[NoFx, R] {
       def apply[A](e: Eff[NoFx, A]) =
         e match {
-          case Pure(a, last) => pure[R, A](a).addLast(last.interpret(apply))
+          case Pure(_, _) => e.asInstanceOf[Eff[R, A]]
           case _ => sys.error("impossible NoFx into R is only for pure values")
         }
     }
@@ -41,24 +41,7 @@ trait IntoPolyLower2  extends IntoPolyLower3 {
   implicit def intoAppendL2L[T1[_], T2[_], R]: IntoPoly[FxAppend[Fx1[T2], R], FxAppend[Fx2[T1, T2], R]] =
     new IntoPoly[FxAppend[Fx1[T2], R], FxAppend[Fx2[T1, T2], R]] {
       def apply[A](e: Eff[FxAppend[Fx1[T2], R], A]): Eff[FxAppend[Fx2[T1, T2], R], A] =
-        e match {
-          case Pure(a, last) =>
-            EffMonad[FxAppend[Fx2[T1, T2], R]].pure(a).addLast(last.interpret(apply))
-
-          case Impure(u@UnionAppendR(r), c, l) =>
-            Impure[FxAppend[Fx2[T1, T2], R], u.X, A](UnionAppendR(r), Arrs.singleton(x => effInto(c(x))), l.interpret(apply))
-
-          case Impure(u@UnionAppendL(Union1(tx)), c, l) =>
-            Impure[FxAppend[Fx2[T1, T2], R], u.X, A](UnionAppendL(Union2R(tx)), Arrs.singleton(x => effInto(c(x))), l.interpret(apply))
-
-          case ImpureAp(unions, continuation, l) =>
-            ImpureAp[FxAppend[Fx2[T1, T2], R], unions.X, A](
-              unions.into(new UnionInto[FxAppend[Fx1[T2], R], FxAppend[Fx2[T1, T2], R]] {
-                def apply[X](union: Union[FxAppend[Fx1[T2], R], X]) = union match {
-                  case UnionAppendR(r) => UnionAppendR(r)
-                  case UnionAppendL(Union1(tx)) => UnionAppendL(Union2R(tx))
-                }}), Arrs.singleton(x => apply(continuation(x))), l.interpret(apply))
-        }
+        e.asInstanceOf[Eff[FxAppend[Fx2[T1, T2], R], A]]
     }
 
   implicit def intoAppendL2R[T1[_], T2[_], R]: IntoPoly[FxAppend[Fx1[T1], R], FxAppend[Fx2[T1, T2], R]] =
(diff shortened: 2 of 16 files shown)


shared/src/main/scala/org/atnos/eff/syntax/either.scala
```diff
@@ -12,7 +12,7 @@ trait either {
 final class EitherEffectOps[R, A](val e: Eff[R, A]) extends AnyVal {
 
   def runEither[E](implicit m: Member[(E Either ?), R]): Eff[m.Out, E Either A] =
-    EitherInterpretation.runEither(e)(m)
+    EitherInterpretation.runEither[R, m.Out, E, A](e)(m)
```
etorreborre: Is this to be more explicit or a type inference regression?

etorreborre:
About type inference for implicit members, there is a test file: [`MemberImplicitsSpec`](https://github.com/atnos-org/eff/blob/master/jvm/src/test/scala/org/atnos/eff/MemberImplicitsSpec.scala). I leave most of it commented out because of compilation times but it can be useful to test regressions.

shared/src/main/scala/org/atnos/eff/Union.scala
```diff
@@ -3,31 +3,50 @@ package org.atnos.eff
...
- *   T2 is in the "middle" of Fx3[T1, T2, T3] so creating a Union object for that effect uses Union3M
+ *  The union type has three concrete constructors:
+ *   - UnionAppendL(nested: Union[L]): Union[FxAppend[L, R]]
+ *   - UnionAppendR(nested: Union[R]): Union[FxAppend[L, R]]
+ *   - UnionTagged(valueUnsafe: Any, index: Int): Union[R] (for R in fx1, fx2, fx3...)
+ *  In that respect UnionTagged behaves similarly to a tagged union in C or C++.
+ *
  */
-sealed trait Union[+R, A] {
-  type X = A
```
etorreborre: Why remove the existential type parameter and replace it with `Any` in various places? Wouldn't we keep a bit more type safety with this `X` because then the following continuation has to take a `u.X` and not anything else in scope?

shared/src/main/scala/org/atnos/eff/Union.scala
```diff
@@ -3,31 +3,50 @@ package org.atnos.eff
...
-sealed trait Union3[R, A] extends Union[R, A]
-case class Union3L[L[_], M[_], R[_], A](t: L[A]) extends Union3[Fx3[L, M, R], A]
-case class Union3M[L[_], M[_], R[_], A](t: M[A]) extends Union3[Fx3[L, M, R], A]
-case class Union3R[L[_], M[_], R[_], A](t: R[A]) extends Union3[Fx3[L, M, R], A]
+  final private[eff] def tagged: UnionTagged[R, A] =
+    this.asInstanceOf[UnionTagged[R, A]]
+}
+case class UnionTagged[R, A] (valueUnsafe: Any, index: Int) extends Union[R, A] {
+  private[eff] def increment[E](i: Int = 1): Union[E, A] = copy(index = index + i)
+  private[eff] def decrement[E](i: Int = 1): Union[E, A] = copy(index = index - i)
```
etorreborre: I don't see where `decrement` and `xor` are being used. Should we keep them?

shared/src/main/scala/org/atnos/eff/Eff.scala
```diff
@@ -296,12 +296,10 @@ trait EffInterpretation {
       case Pure(a, Last(None))    => Monad[M].pure(Right(a))
 
       case Impure(u, continuation, last) =>
-        u match {
-          case Union1(ta) =>
-            last match {
-              case Last(Some(l)) => ta.map(x => Left(continuation(x).addLast(last)))
-              case Last(None)    => ta.map(x => Left(continuation(x)))
-            }
+        val ta = u.tagged.valueUnsafe.asInstanceOf[M[A]]
```
etorreborre: `.tagged` is an unsafe operation. At this stage how do you know that you have a `UnionTagged` and not a `UnionAppend`?

shared/src/main/scala/org/atnos/eff/IntoPoly.scala
```diff
@@ -26,7 +26,7 @@ trait IntoPolyLower1 extends IntoPolyLower2 {
     new IntoPoly[NoFx, R] {
       def apply[A](e: Eff[NoFx, A]) =
         e match {
-          case Pure(a, last) => pure[R, A](a).addLast(last.interpret(apply))
+          case Pure(_, _) => e.asInstanceOf[Eff[R, A]]
```
etorreborre: I was not sure that the cast was safe here, but I think it is because `last` can only be an `Option[Eval[Pure[Unit]]`.

shared/src/main/scala/org/atnos/eff/IntoPoly.scala
```diff
@@ -41,99 +41,65 @@ trait IntoPolyLower2  extends IntoPolyLower3 {
...
     new IntoPoly[FxAppend[Fx1[T1], R], FxAppend[Fx2[T1, T2], R]] {
       def apply[A](e: Eff[FxAppend[Fx1[T1], R], A]): Eff[FxAppend[Fx2[T1, T2], R], A] =
         e match {
           case Pure(a, last) =>
-            EffMonad[FxAppend[Fx2[T1, T2], R]].pure(a).addLast(last.interpret(apply))
+            pure(a).addLast(last.interpret(apply))
 
           case Impure(u@UnionAppendR(r), c, l) =>
-            Impure[FxAppend[Fx2[T1, T2], R], u.X, A](UnionAppendR(r), Arrs.singleton(x => effInto(c(x))), l.interpret(apply))
+            Impure[FxAppend[Fx2[T1, T2], R], Any, A](Union.appendR(r), Arrs.singleton(x => apply(c(x.asInstanceOf[Nothing]))), l.interpret(apply))
```
etorreborre: Would the `x.asInstanceOf[Nothing]` be removed if we kept `u.X`?

shared/src/main/scala/org/atnos/eff/syntax/either.scala
edmundnoble: To be more explicit.

shared/src/main/scala/org/atnos/eff/Union.scala
edmundnoble: This is totally my bad, I tried going on a crusade to make everything covariant and that demanded dropping the `X` type member. When I went back I didn't go all the way back. Thanks for pointing it out :smile:

shared/src/main/scala/org/atnos/eff/Union.scala
edmundnoble: They're used in `Member.scala`.

shared/src/main/scala/org/atnos/eff/Eff.scala
edmundnoble: Because `R =:= Fx1[M]`.

shared/src/main/scala/org/atnos/eff/IntoPoly.scala
edmundnoble: :+1:

shared/src/main/scala/org/atnos/eff/IntoPoly.scala
edmundnoble: Absotively, it will be reverted when I add `X` back.

edmundnoble:
Comments hopefully addressed :)

shared/src/main/scala/org/atnos/eff/Eff.scala
etorreborre: Right, I didn't expand the code enough to see that, thanks.

shared/src/main/scala/org/atnos/eff/Member.scala
```diff
@@ -3,224 +3,162 @@ package org.atnos.eff
...
 
-  def transform[O[_]](implicit from: O ~> T): MemberIn[O, R] = new MemberIn[O, R] {
+  final def transform[O[_]](implicit from: O ~> T): MemberIn[O, R] = new MemberIn[O, R] {
     def inject[V](ov: O[V]): Union[R, V] =
       outer.inject(from(ov))
   }
 }
 
+final case class TaggedMemberIn[T[_], R](tag: Int) extends MemberIn[T, R] {
+  override def inject[V](tv: T[V]): Union[R, V] = UnionTagged(tv, tag)
```
etorreborre: Minor point but the `override` is still superfluous here, right?

shared/src/main/scala/org/atnos/eff/Member.scala
```diff
@@ -371,23 +293,23 @@ trait MemberLower6 extends MemberLower7 {
...
-    def project[V](union: Union[FxAppend[Fx1[T], Fx3[L, M, R]], V]): Union[Fx3[T, L, R], V] Either M[V] =
-      union match {
-        case UnionAppendL(Union1(t))  => Left(Union3L(t))
-        case UnionAppendR(Union3L(l)) => Left(Union3M(l))
-        case UnionAppendR(Union3M(m)) => Right(m)
-        case UnionAppendR(Union3R(r)) => Left(Union3R(r))
-        case UnionAppendR(_)          => sys.error("appeasing the compiler exhaustiveness checking for Member4RM")
+    def project[V](union: Union[FxAppend[Fx1[T], Fx3[L, M, R]], V]): Union[Fx3[T, L, R], V] Either M[V] = union match {
+      case UnionAppendL(tagged@UnionTagged(_, _)) => Left(tagged.copy(index = 3))
+      case UnionAppendR(tagged@UnionTagged(value, index)) => (index: @switch) match {
```
etorreborre: Nice use of `@switch` I forgot about that.

shared/src/main/scala/org/atnos/eff/Member.scala
```diff
@@ -396,41 +318,37 @@ trait MemberLower7 extends MemberLower8 {
...
-    def accept[V](union: Union[Fx3[T, L, M], V]): Union[FxAppend[Fx1[T], Fx3[L, M, R]], V] =
-      union match {
-        case Union3L(t) => UnionAppendL(Union1(t))
-        case Union3M(l) => UnionAppendR(Union3L(l))
-        case Union3R(m) => UnionAppendR(Union3M(m))
-      }
+    def accept[V](union: Union[Fx3[T, L, M], V]): Union[FxAppend[Fx1[T], Fx3[L, M, R]], V] = union.tagged match {
+      case tagged@UnionTagged(_, index) => 
+        if (index == 3) UnionAppendL(tagged.copy(index = 1))
+        else UnionAppendR(tagged.increment())
```
etorreborre: Also minor, I usually provide 2 versions of these kind of methods to avoid the awkward (IMO) `()`:
```scala
def increment: UnionTagged[R, A]
def increment(i: Int) = UnionTagged[R, A]
```

shared/src/main/scala/org/atnos/eff/Member.scala
```diff
@@ -371,23 +293,23 @@ trait MemberLower6 extends MemberLower7 {
...
 
-    def accept[V](union: Union[Fx3[T, L, R], V]): Union[FxAppend[Fx1[T], Fx3[L, M, R]], V] =
-      union match {
-        case Union3L(t) => UnionAppendL(Union1(t))
-        case Union3M(l) => UnionAppendR(Union3L(l))
-        case Union3R(r) => UnionAppendR(Union3R(r))
+    def accept[V](union: Union[Fx3[T, L, R], V]): Union[FxAppend[Fx1[T], Fx3[L, M, R]], V] = union.tagged match {
+      case tagged@UnionTagged(_, index) => (index: @switch) match {
+        case 1 => UnionAppendR(tagged.forget)
+        case _ => UnionAppendL(tagged.xor())
```
etorreborre: If `index = 2` (for a `L` effect) then using `xor` makes `2 ^ 2 == 0` and we end-up with a `UnionTagged(_, 0)` for the nested union inside `UnionAppendL`. But all indices for `UnionTagged` should always be `1, 2 or 3`. Am I reading this wrong? I would feel more confident being explicit here or instead make sure we prove why it is correct. This part of the `eff` code worries me a lot because it has a huge potential for runtime errors.

edmundnoble: Very good point, it looks like I was refactoring my attempt to make Append part of the indexing as well. I'll run those `MemberImplicitSpec` regression tests to see if there's anything else like this lurking there.

shared/src/main/scala/org/atnos/eff/Member.scala
edmundnoble: Turns out I could remove all the arguments and just use 1 :D

shared/src/main/scala/org/atnos/eff/Member.scala
edmundnoble: Removed.

shared/src/main/scala/org/atnos/eff/Member.scala
edmundnoble: :+1: (hopefully it's faster)

shared/src/main/scala/org/atnos/eff/Member.scala
etorreborre: I am not sure the tests in `MemberImplicitsSpec` will catch this issue because the commented out tests are more about implicits inference. In order to exercice this `accept` capability you would have to use something like `transform` with a specific stack I think.

edmundnoble: Today I'll look at extending `MemberSpec` with more concrete `Stack`s.

edmundnoble: Added it. The accept/project/inject law handles these methods.

etorreborre:
👍

PR #60, opened 2017-02-02
Added a method to update the read value of a Reader effect

etorreborre:
@benhutchison @edmundnoble we're a bit short on names for all the ways of modifying an effect with 2 type parameters but I think this is useful (cf the recent question on Gitter).

### DIFF
--- a/jvm/src/test/scala/org/atnos/eff/EitherEffectSpec.scala
+++ b/jvm/src/test/scala/org/atnos/eff/EitherEffectSpec.scala
@@ -137,7 +137,7 @@ class EitherEffectSpec extends Specification with ScalaCheck { def is = s2"""
       EitherEffect.left(Error1("boom"))
 
     val action2: Eff[R2, Unit] =
-      action1.localEither(Error2)
+      action1.zoomEither(Error2)
 
     action2.runEither.run ==== Left(Error2(Error1("boom")))
   }
--- a/jvm/src/test/scala/org/atnos/eff/EitherEffectSpec.scala
+++ b/jvm/src/test/scala/org/atnos/eff/EitherEffectSpec.scala
@@ -151,7 +151,7 @@ class EitherEffectSpec extends Specification with ScalaCheck { def is = s2"""
     val action1: Eff[R1, Unit] =
       EitherEffect.left(Error1("boom"))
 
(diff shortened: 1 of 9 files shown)


edmundnoble:
I'm not sure, I think that this really should be called `localReader` if we add it. Might a name like `zoomReader` be more appropriate for the current `localReader`?

acieroid:
For the name, `localReader` would make more sense I think. For example, Haskell's reader monad has `local :: (r -> r) -> m a -> m a` that corresonds to that method.
I also think that the current `localReader` of Eff would make more sense as `zoomReader`.

etorreborre:
When I wrote the first `localReader` I thought of it as a generalized version of Haskell's `local`. I agree that changing names and using `zoom` is better. I will do it. I am just annoyed by the API break that this entails (there's a similar method on the Either effect), but since the `eff` community is probably not yet very large, I propose that we fold this and all the recent changes into a `3.0.0`.

etorreborre:
We might as well yes

benhutchison:
Personally Im OK with breaking source compatibility at present. Eff codebases are few, young and there's still a need to refine and experiment the framework.

etorreborre:
@benhutchison it's good to know there's a green light on your side :-).

etorreborre:
@edmundnoble @acieroid I have added a fixup commit containing:

 - a change to name `local` when the type of the read value doesn't change, `zoom` when we "replace" a reader of a given type with a reader of another type, `runLocal` when we interpreter a `Reader` effect into another one in the remaining spec (I don't think this one is really useful though)

 - the same naming applied to `either` 
 - a `local` method for `State`

edmundnoble:
@etorreborre Perhaps `runReaderWithZoomedReader` more clearly expresses that `runLocalReader` does? Not really sure, it is quite a long name, but the fact that it eliminates one reader using another "zoomed out" reader (in the same stack) isn't incredibly obvious from the name.

etorreborre:
I would really prefer a shorter name. I was also thinking at just `translateReader` since the underlying `Interpreter.translate` method is being used. In that case we "translate" a Reader effect into another one in the case (what else could we translate it too anyway?)

edmundnoble:
Yeah that's alright with me as well.

etorreborre:
Great, I just pushed a fixed and I'm going to sleep on it. Goodnight!

PR #61, opened 2017-02-04
Cookbook page for usage examples, with example of partial interpretation

benhutchison:


etorreborre:
Thanks @benhutchison, I took your branch and incorporated it with some edits.

PR #62, opened 2017-02-05
Implement detach in terms of detachA

edmundnoble:
Beside removing code, this removes a monad bind per `ImpureAp`.


etorreborre:
👍

PR #65, opened 2017-02-06
Release 3.0.0

etorreborre:
@edmundnoble I removed the `Async` effect but also moved the implicit for `TaskOps` to a different object to make sure it is really imported when required (and not when importing `TaskEffect`). This is more consistent with the rest of the other effects (and good practice I think).

I also added the notes for that new release full of improvements thanks to you and @benhutchison!


edmundnoble:
:+1: I was considering looking into the imports as well :)

etorreborre:
Thanks I will publish tomorrow (or maybe the day after, tomorrow might be busy).

PR #68, opened 2017-02-14
Added a 'onNone' action on Arrs continuation to clean up resources

etorreborre:
This is used to allow the `finally` interceptors in the Safe effect to be executed even if an other effect short-circuits all computations (see the `bracket` test cases in the `SafeEffectSpec`).

I have also added a `bracketLast` method to `Eff`. The main difference with `SafeEffect.bracket` are:
 - that the "last" action is always executed at the end of the program whether or not the protected Eff computation has additional flatMaps or not

 - no finalizer exception is reported with the "last" action

**Implementation**

The `Arrs` class gets an additional `onNone` parameter representing an action to execute when there is no value available to feed the continuation. This is the case when the `Safe` effect wants to protect computations and clean-up resources but an additional `Either` effect stops all computations with a `Left` value. This issue happened on the [eff gitter channel](https://gitter.im/atnos-org/eff?at=58a175ee00c00c3d4f301675) first.

@edmundnoble this is a massive PR where most of the work is in making sure that the `onNone` action doesn't get lost. There's probably a missing `Profunctor` instance lost for `Arrs` but it's a bit too late for me to fix this :-).

### DIFF
--- a/shared/src/main/scala/org/atnos/eff/Arrs.scala
+++ b/shared/src/main/scala/org/atnos/eff/Arrs.scala
@@ -11,8 +11,12 @@ import scala.annotation.tailrec
  *
  *  A => Eff[R, X1]; X1 => Eff[R, X2]; X2 => Eff[R, X3]; ...; X3 => Eff[R, B]
  *
+ * An alternate unit value can also be set on this function in case the argument A is not available.
+ * This value can be set by an effect to do some cleanup if it doesn't even get the chance
+ * to add its own effect. See SafeEffect.bracket
+ *
  */
-case class Arrs[R, A, B](functions: Vector[Any => Eff[R, Any]]) extends (A => Eff[R, B]) {
+case class Arrs[R, A, B](functions: Vector[Any => Eff[R, Any]], onNone: Last[R] = Last.none[R]) extends (A => Eff[R, B]) {
 
   /**
    * append a new monadic function to this list of functions such that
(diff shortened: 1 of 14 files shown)


shared/src/main/scala/org/atnos/eff/Arrs.scala
```diff
@@ -69,20 +73,32 @@ case class Arrs[R, A, B](functions: Vector[Any => Eff[R, Any]]) extends (A => Ef
...
   def contramap[C](f: C => A): Arrs[R, C, B] =
-    Arrs(((c: Any) => Eff.pure[R, Any](f(c.asInstanceOf[C]).asInstanceOf[Any])) +: functions)
+    Arrs(((c: Any) => Eff.pure[R, Any](f(c.asInstanceOf[C]).asInstanceOf[Any])) +: functions, onNone)
 
   def contraFlatMap[C](f: C => Eff[R, A]): Arrs[R, C, B] =
-    Arrs(f.asInstanceOf[Any => Eff[R, Any]] +: functions)
+    Arrs(f.asInstanceOf[Any => Eff[R, Any]] +: functions, onNone)
+
+  /** adapt the input and output of an Arrs function */
+  def dimap[C, D](f: C => A)(g: Eff[R, B] => Eff[R, D]): Arrs[R, C, D] =
```
edmundnoble: Sorry, but this ain't `dimap` ;) `g` has to be `B => D` for it to be `dimap`. No reason not to include this though: maybe `dimapEff`?

edmundnoble:
Looks mostly good to me. `Last` is one of those areas that seems easy to get wrong, so I'm glad we're adding tests :)

shared/src/main/scala/org/atnos/eff/Arrs.scala
etorreborre: Right, it's not really `dimap`.

benhutchison:
@etorreborre Thanks a heap for putting this together so quickly! :)

wjlow:
This is awesome. Thanks so much!

PR #69, opened 2017-02-14
Fix State.modify for applicative case

edmundnoble:
Amazingly, State.modify does not work properly if used applicatively. All of the state updates get run one after the other, but they don't get the results of the previous one.


etorreborre:
I think this is [the right thing to do](https://gitter.im/atnos-org/eff?at=58a3321600c00c3d4f3a6e1d).

PR #73, opened 2017-03-04
 Replace ScheduledExecutorService in async effects with Timer for JS compat

edmundnoble:
Users on JS cannot manufacture `java.util.concurrent.ScheduledExecutorService`s with which to call `runAsync` or `runSequential` because it is not included in their version of the Java stdlib. This PR thus replaces all of their occurrences with `java.util.Timer`.


edmundnoble:
Closing in favor of #74.

PR #76, opened 2017-03-12
Add pure union

etorreborre:
@edmundnoble please have a look at this PR. It started from a use case at work where we had a SOE with a large chain of flatMaps just using `pure`. 

In this PR I introduce the possibility to have a pure value in the `Impure` case, denoted as `NoEffect(a)` instead of the usual `Union` of effects. 

So now we have:
```
sealed trait Effect[R, A]
case class NoEffect[R, A](a: A) extends Effect[R, A]
sealed trait Union[R, A] extends Effect[R, A]
```

This lets me use a continuation even on pure values, and avoid the issue with flatMaps on pure values.

More importantly I think this opens the possibility to write a generic interpreter like `handle_relay` in the original paper. I wrote a small prototype for the `Writer` effect and it works. This could greatly simplify the `Interpret` trait (at the expense of me rewriting lots of stuff...)

### DIFF
--- a/jvm/src/test/scala/org/atnos/eff/EffSpec.scala
+++ b/jvm/src/test/scala/org/atnos/eff/EffSpec.scala
@@ -29,6 +29,7 @@ class EffSpec extends Specification with ScalaCheck { def is = s2"""
  The Eff monad is stack safe with Reader                 $stacksafeReader
  The Eff monad is stack safe with both Reader and Writer $stacksafeReaderWriter
  The Eff monad is tail recursive                         $tailrec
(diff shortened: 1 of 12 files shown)


edmundnoble:
Is `traverseA` stack-safe safe now?

etorreborre:
@edmundnoble no it's still not stacksafe. I think it could be but it's a bit out of my reach right now. I think the main issue is in the definition of `effApplicativeUnsafe.ap` which is not tail-recursive and which might be if we used some clever stack. But that will be for another PR.

PR #77, opened 2017-03-13
Implement update in tutorial to include call to put 

mdr:
Mirroring cats tutorial implementation: http://typelevel.org/cats/datatypes/freemonad.html


etorreborre:
Thanks for the PR. Since these html pages are generated I also did the change on the corresponding Scala code in 6d75436

PR #78, opened 2017-03-15
update catbird and twitter util

crispywalrus:
not sure how sensitive twitter util is wrt binary compatibility but
since catbird brings in both util and finagle I've upgraded this to
current versions.


etorreborre:
Thanks for the update!

PR #79, opened 2017-03-16
Add future syntax to all

Unisay:


etorreborre:
I did not add it to `all` on purpose. Because it is the same syntax for other `task` effects and will conflict if you have both effects. So it's better to do an explicit import.

PR #81, opened 2017-03-20
Tail recursive generic interpreter

etorreborre:
@edmundnoble this is a big PR sorry.

The main idea is to use the previous `Impure(NoEffect(x), continuation)` functionality to (finally) implement a generic stack-safe interpreter just taking in parameter functions declaring what to do with a pure value, an effect, a bunch of effects.

I have also done some renamings like `Arrs` to `Continuation`, departing from the original article because I think it is more understandable for newcomers. 

I will probably add a bit more doc on how to write interpreters (now that it is simpler) in the next few days, then I want to publish a `eff-4.0.0-RC1` for people to try especially because there are low-level API changes.

cc @benhutchison @stephanh @reactormonk


jvm/src/test/scala/org/atnos/eff/WriterEffectSpec.scala
```diff
@@ -38,11 +39,11 @@ class WriterEffectSpec extends Specification { def is = s2"""
     val action: Eff[S, String] = for {
       f <- tell[S, String]("hello")
       h <- tell[S, String]("world")
-    } yield f+" "+h
+    } yield "hello world"
```
edmundnoble: I do not understand the rationale for this change, nor the one at line 25 above.

shared/src/main/scala/org/atnos/eff/Eff.scala
```diff
@@ -291,27 +299,40 @@ trait EffCreation {
   def whenStopped[R, A](e: Eff[R, A], action: Last[R]): Eff[R, A] =
     e match {
       case Pure(a, l)        => Pure(a, l)
-      case Impure(u, c, l)   => Impure(u, Arrs.singleton((x: u.X) => whenStopped[R, A](c(x), action), action), l)
-      case ImpureAp(u, c, l) => ImpureAp(u, Arrs.singleton((x: Vector[Any]) => whenStopped[R, A](c(x), action), action), l)
+      case Impure(u, c, l)   => Impure(u,   c.copy(onNone = c.onNone <* action), l)
+      case ImpureAp(u, c, l) => ImpureAp(u, c.copy(onNone = c.onNone <* action), l)
```
edmundnoble: :+1:

shared/src/main/scala/org/atnos/eff/Eff.scala
```diff
@@ -257,9 +257,17 @@ trait EffCreation {
     Pure(a)
 
   /** create a impure value from an union of effects and a continuation */
-  def impure[R, X, A](union: Union[R, X], continuation: Arrs[R, X, A]): Eff[R, A] =
+  def impure[R, X, A](union: Union[R, X], continuation: Continuation[R, X, A]): Eff[R, A] =
     Impure[R, X, A](union, continuation)
 
+  /** create a delayed impure value */
+  def impure[R, A, B](value: A, continuation: Continuation[R, A, B]): Eff[R, B] =
```
edmundnoble: What... what does this do? Should this be an `: => A`?

shared/src/main/scala/org/atnos/eff/IntoPoly.scala
```diff
@@ -15,7 +15,10 @@ import Eff._
  * Note: some implicit definitions are probably missing in some cases
  */
 trait IntoPoly[R, U] {
-  def apply[A](e: Eff[R, A]): Eff[U, A]
+  val unionInto: UnionInto[R, U]
```
edmundnoble: :+1: I was considering making an issue on this; the fact that `IntoPoly` can't really have inductive instances. Nice to see it fixed.

shared/src/main/scala/org/atnos/eff/ListEffect.scala
```diff
@@ -41,69 +41,28 @@ trait ListCreation {
...
-      def onPure(a: A, s: S): (Eff[R, A], S) Either Eff[U, List[A]] =
-        s match {
-          case (head :: tail, result) => Left((head, (tail, result :+ a)))
-          case (List(), result)       => Right(EffMonad[U].pure((result :+ a).toList))
+  def runList[R, U, A](effect: Eff[R, A])(implicit m: Member.Aux[List, R, U]): Eff[U, List[A]] =
+    runInterpreter(effect)(new Interpreter[List, U, A, List[A]] {
+      def onPure(a: A): Eff[U, List[A]] =
+        Eff.pure(List(a))
+
+      def onEffect[X](xs: List[X], continuation: Continuation[U, X, List[A]]): Eff[U, List[A]] =
```
edmundnoble: Can we make a default interpreter for everyone with a Traverse instance?

shared/src/main/scala/org/atnos/eff/WriterEffect.scala
```diff
@@ -46,68 +46,79 @@ trait WriterInterpretation {
...
-    def fold(a: A, s: S) = s >> f(a)
+    val init = Eval.later(())
+    def fold(a: A, s: S) = Eval.later { f(a) >> s }.flatten
     def finalize(s: S) = s
   }
 
 }
 
 /** support trait for folding values while possibly keeping some internal state */
-trait LeftFold[A, B] {
```
edmundnoble: :+1:

shared/src/main/scala/org/atnos/eff/Eff.scala
edmundnoble: NVM, I see :)

shared/src/main/scala/org/atnos/eff/ListEffect.scala
etorreborre: That's a good idea. We could even have a `TraverseEffect` to generalize the `ListEffect`.

Issue #80, opened 2017-03-19
Help to convert tutorials to running programs within a project


russellcameronthomas:
I am relatively new to Scala (with Java background) and I am teaching myself Functional Programming.  I am working my way through your tutorials.  

My problem is that I am only able to get them running through the command line REPL.  I have been trying to take the (simple) step of inserting each of the tutorials into a file within _IntelliJ IDEA_ so that they print results to the console.  My problem is they won't compile without errors, and I don't know why.

For example, taking code from the "Introduction" here is the file I have created:
```
class Tutorial{

}

object Intro extends App {
  import cats._
  import cats.data._
  import org.atnos.eff._

  type ReaderInt[A] = Reader[Int, A]
  type WriterString[A] = Writer[String, A]

  type Stack = Fx.fx3[WriterString, ReaderInt, Eval]

  import org.atnos.eff.all._
  import org.atnos.eff.syntax.all._

  // useful type aliases showing that the ReaderInt and the WriterString effects are "members" of R
  // note that R could have more effects
  type _readerInt[R]    = ReaderInt |= R
  type _writerString[R] = WriterString |= R


  def program[R: _readerInt : _writerString : _eval]: Eff[R, Int] = for {
  // get the configuration
    n <- ask[R, Int]

    // log the current configuration value
    _ <- tell("the required power is " + n)

    // compute the nth power of 2
    a <- delay(math.pow(2, n.toDouble).toInt)

    // log the result
    _ <- tell("the result is " + a)
  } yield a


  println(program[Stack].runReader(6).runWriter.runEval.run)
}

```

The compiler error is in the last line: "Cannot resolve symbol 'run'".   I have tried many other approaches and structures, and they all give this error, or some other "Cannot resolve..." error.

I have already tried the "clear cache" fix, with restart of _IntelliJ IDEA_.

Can anyone help?

reactormonk:
This looks more like a stackoverflow question, could you post the link to the question here?

etorreborre:
There are 2 compiler plugins which are necessary for compilation:
```
// to write types like Reader[String, ?]
addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.9.3")

// to get types like Reader[String, ?] (with more than one type parameter) correctly inferred
// this plugin is not necessary with Scala 2.12
addCompilerPlugin("com.milessabin" % "si2712fix-plugin_2.11.8" % "1.2.0")
```

If you want to compile with Intellij you probably have to specify those plugins:
[...]

PR #82, opened 2017-03-20
Updated installation fallback & intellij notice

reactormonk:
In reference to https://github.com/atnos-org/eff/issues/80

### DIFF
--- a/jvm/src/test/scala/org/atnos/site/Installation.scala
+++ b/jvm/src/test/scala/org/atnos/site/Installation.scala
@@ -9,10 +9,27 @@ libraryDependencies += "org.atnos" %% "eff" % "$version"
 
 // to write types like Reader[String, ?]
 addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.9.3")
-
-// to get types like Reader[String, ?] (with more than one type parameter) correctly inferred
-addCompilerPlugin("com.milessabin" % "si2712fix-plugin_2.11.8" % "1.2.0")
 ```
+
+To get types like Reader[String, ?] (with more than one type parameter) correctly inferred, you'll need scala 2.12.1 with
+
+```scala
+scalacOptions += "-Ypartial-unification"
+```
+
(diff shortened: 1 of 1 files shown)


jvm/src/test/scala/org/atnos/site/Installation.scala
```diff
@@ -9,11 +9,28 @@ libraryDependencies += "org.atnos" %% "eff" % "$version"
 
 // to write types like Reader[String, ?]
 addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.9.3")
+```
+
+To get types like Reader[String, ?] (with more than one type parameter) correctly inferred, you'll need scala 2.12.0 with
```
etorreborre: s/2.12.0/2.12.1

jvm/src/test/scala/org/atnos/site/Installation.scala
```diff
@@ -9,11 +9,28 @@ libraryDependencies += "org.atnos" %% "eff" % "$version"
...
 addCompilerPlugin("com.milessabin" % "si2712fix-plugin_2.11.8" % "1.2.0")
 ```
 
+or you could use
+
+```scala
+scalaOrganization := "org.typelevel"
+```
+
+to get the `-Ypartial-unification` flag for scala 2.11.8
```
etorreborre: Nice!

jvm/src/test/scala/org/atnos/site/Installation.scala
```diff
@@ -70,6 +87,10 @@ You can also access all the syntax imports at once with:
 import org.atnos.eff.syntax.all._
 ```
 
+#### Intellij support
+
+Intellij doesn't support implicit-directed type inference. It'll work when you build it.
```
etorreborre: I would like to avoid snarky comments here :-). Can you please reference issues like https://youtrack.jetbrains.com/issue/SCL-11140 or https://youtrack.jetbrains.com/issue/SCL-10753 instead?

etorreborre:
Thanks @reactormonk!

PR #83, opened 2017-03-21
Generate boilerplate using macros

nightscape:
**This is still WIP**

We are doing our first steps with free monads and have previously been using the corresponding cats implementation. To reduce the amount of required boilerplate, we used the very nice [Freasy-Monad library](https://github.com/Thangiee/Freasy-Monad) on top of cats.

As we're currently evaluating switching over to eff (mostly because stack transformations seem much nicer) and we liked the "macro generates the boilerplate" approach, we ported the Freasy-Monad (MIT licensed) code to work with eff.

The macro reads the method definitions from a `@eff trait` and generates the corresponding ADT `case classes`, a `SideEffect`, `Translate` and a `FunctionK` sub-trait which the user then has to fill with methods that resemble the methods from the `@eff trait`.

The tests resemble the tutorial and show what the user still needs to implement.

We're happy about any feedback!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .idea
 target/
 .settings
+/.ensime*
--- a/build.sbt
+++ b/build.sbt
@@ -11,8 +11,8 @@ lazy val eff = project.in(file("."))
   .settings(moduleName := "root")
   .settings(effSettings)
   .settings(noPublishSettings)
-  .aggregate(coreJVM, coreJS, monixJVM, monixJS, scalaz, twitter, fs2JS, fs2JVM)
-  .dependsOn(coreJVM, coreJS, monixJVM, monixJS, scalaz, twitter, fs2JS, fs2JVM)
+  .aggregate(coreJVM, coreJS, macros, monixJVM, monixJS, scalaz, twitter, fs2JS, fs2JVM)
+  .dependsOn(coreJVM, coreJS, macros, monixJVM, monixJS, scalaz, twitter, fs2JS, fs2JVM)
 
 lazy val core = crossProject.crossType(CrossType.Full).in(file("."))
   .settings(moduleName := "eff")
@@ -24,6 +24,14 @@ lazy val core = crossProject.crossType(CrossType.Full).in(file("."))
 
 lazy val coreJVM = core.jvm
 lazy val coreJS = core.js
+
+lazy val macros = project.in(file("macros"))
+  .settings(moduleName := "eff-macros")
+  .dependsOn(coreJVM)
+  .settings(libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value)
+  .settings(addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full))
+  .settings(commonJvmSettings)
+  .settings(effSettings:_*)
 
 lazy val monix = crossProject.crossType(CrossType.Full).in(file("monix"))
   .settings(moduleName := "eff-monix")
--- a/macros/src/main/scala/org/atnos/eff/macros/EffMacros.scala
+++ b/macros/src/main/scala/org/atnos/eff/macros/EffMacros.scala
@@ -166,7 +166,7 @@ class EffMacros(val c: blackbox.Context) {
                       cq"${adt(name)} => $name"
                   }}
                 }
-                def applicative[X, Tr[_] : Traverse](ms: Tr[${sealedTrait.name}[X]]): Tr[X] =
+                def applicative[X, Tr[_] : cats.Traverse](ms: Tr[${sealedTrait.name}[X]]): Tr[X] =
                   ms.map(apply)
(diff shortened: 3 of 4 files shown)


nightscape:
Btw, the `old_macro` branch name stems from the fact that we also tried Scala.meta paradise. The Scala.meta implementation is nicer, but unfortunately leads to a compiler crash due to a `Position error: Overlapping regions`.

etorreborre:
Thanks for this! I don't create effects every day because I'm mostly using eff as a replacement for MTL but I understand that many people use it for DSLs. So I think it is indeed a good idea to have some macro support.

nightscape:
Great! Would you mind to check if the generated code makes sense (it currently still gets `println`ed when running tests)? I'm only beginning to learn Eff and there might be some issues lurking in there..

etorreborre:
I just had a quick look at the code. Some comments:

 - I have seen some issues in the past with the compiler crashing when the ADT for the effect was nested inside a trait. But maybe we can hope that this will not be the case anymore?

 - I'm not sure that providing a `SideEffect` by default is a nice service to make to users. It can be useful in some rare cases but should be avoided in general

 - I would generate everything in the `KVStore` trait then just have `object KVStore extends KVStore` if people want to stack traits.

 - I don't know why the generated code shows one `()` value at the end. I'm not sure if that can be a problem with some scala warning which would be warning for an unused value.

 - all the rest makes sense to me

nightscape:
Hi Eric, thanks for the review!

I made the macro generate a `KVStoreDsl` containing the code and a corresponding companion object.
I also flattened the structure of the generated code so that the both the ADTs and the helper methods that generate them are now directly within the `KVStoreDsl`.
That has the additional benefit that the generated `KVStoreDsl` now is much more similar to the user-specified `KVStoreDsl`, especially that the methods are directly inside the trait, instead of an `object ops { ... }`.
I also don't know about the generated `()`, but it doesn't seem to cause problems. Are macros even checked for warnings?
About the `SideEffect` trait: We're using it mainly to mock out interpreters for testing, but there it is very helpful :smiley:

Another open question (that would allow me to remove some code):
Is it possible to also have "helper `val`s"? As in
```scala
val someConst: Eff[R, Option[T]]
```
I would assume it isn't, because you would not be able to specify the effect stack `R`, except if one maybe provided `R` on another level or would have some kind of "empty stack" as R.

Have you ever seen a `val` like that? Could you imagine that it might be needed?

edmundnoble:
This is very nice; few concerns:

1. To clarify, this is an optional dependency, right? `eff-macros`? Just asking because a lot of people using Eff do not need this at all; specifying DSLs in final tagless style for example obviates the need for the entire thing (see https://github.com/edmundnoble/final-tagless-typelevel-summit) and actually requires the same user code as your macro does :)

2. I appreciate what you're saying about SideEffect, that tests need easy generation, but I still do not think it should be in the companion object. The generated FunctionK is just as easy to use for this, and less likely to be misused.

3. I'd suggest considering using dynamic dispatch to run your FunctionK and Translate for performance reasons; but honestly I'm not sure how many cases your algebra would need to have to make a difference. That is, give each concrete subclass of the `@eff` trait a `runFunctionK` and `runTranslate` method that calls the exact method on the generated `FunctionK` and `Translate` subtraits that it corresponds to, so that there's no need to pattern match.

P.S. you can use `NoFx` for an empty stack; you will be doing `into` calls instead of `pure` calls when it's used though, so it's not a huge win IMO.

nightscape:
Hi Edmund, thanks for your comments!

1. This would be an optional package, yes. I only stumbled upon your final tagless talk after having written the macro stuff. I still don't understand how final tagless and eff play together. Under which circumstances would you recommend the final tagless style over extensible effects?
2. I had used the `FunctionK` instances before, but that yielded stacks like `FxAppend[Fx1[Task],FxAppend[Fx1[Task],Fx3[Task,Task,Task]]]` (`Task` being `monad.eval.Task`) and I don't know how to reduce that. I imagine the same would apply for `Id` instead of `Task`? Would you not generate `SideEffect` at all or just hide it a little more, e.g. in an `object NotSureIf`? :wink:
3. That would probably work and is a great idea for `FunctionK`, but I have no clue how to generate the `runTranslate` method as AFAIU it would need implicits for all the effects in `U` that the resulting `Eff[U, ...]` will have, and those are only know at the time when the sub-class is implemented. I guess one could write another macro like `@translates(classOf[KVStoreDsl], SomeRepresentationOfTheResultingEffects)` which is applied to the user-implemented class. Do you know of an easier way?

nightscape:
The new commit might explain it: There's now `TranslatorFactory1` ... `TranslatorFactory9` where the number determines how many "outgoing" effects the "incoming" effect is translated to.

edmundnoble:
Hey @nightscape, very welcome:

1. Domain-specific algebras, especially ones that will be allocated a lot during your program, can be replaced with finally tagless for an efficiency boost (and be much easier to understand, IMO).

2. I think that is more of an issue with Eff. We need a mechanism for interpreting several effects at once, which could do double-duty as collapsing several identical effects.

3. Nicely done :D

etorreborre:
@nightscape time to merge and let the rubber hit the road!

PR #84, opened 2017-03-29
>>= alias for flatMap

Unisay:


etorreborre:
Why not use `import cats.syntax.flatMap._`?

Unisay:
I have somehow overlooked it, was searching in `eff.syntax`

PR #85, opened 2017-03-30
Changed the smart constructor names for the KVStore effect

etorreborre:


edmundnoble:
What precedent is there for this exactly? Forgive me, I'm not up to date on the modern KV-store command-sets.

etorreborre:
I am just answering @russellcameronthomas issue with the tutorial which is introducing conflicting names for the KVStore effect and the State effect. Otherwise I don't know if there's a blessed set of command names for KV stores.

edmundnoble:
Oh yes, I see. :+1:

PR #87, opened 2017-04-02
Bring Translate docs closer to Transform docs

edmundnoble:
Also fill in some `???`.


jvm/src/test/scala/org/atnos/site/TransformStack.scala
```diff
@@ -164,19 +166,22 @@ def runAuth[R, U, A](e: Eff[R, A])(implicit
...
            collapse
        }
     })
 
 // call to a service to authenticate tokens
-def authenticate(token: String): TimedFuture[AuthError Either AccessRights] = ???
+def authenticateImpl(token: String): Future[AuthError Either AccessRights] =
+  Future.successful[AuthError Either AccessRights] { Left(AuthError("token invalid!")) }
+
+def authenticate[S: _authenticate](token: String) = Authenticate(token).send
```
etorreborre: Just one thing about formatting. Everywhere in the code I have written:
```scala
def authenticate[R :_authenticate]
```
The idea is to be close to a "regular" type bound but with a `_`.

edmundnoble: Ah yes my mistake. scalafmt, wherefore art thou

etorreborre:
Thanks!

PR #88, opened 2017-04-03
Detach with last

etorreborre:
@edmundnoble `MonadError` needs to be used with `detach` in order to make sure that `Last` actions are indeed being triggered when `Future` or `Task` are failing.


edmundnoble:
For the record; I don't like that we have this requirement, but it makes perfect sense to me.

etorreborre:
I'm not a big fan either but I don't see any obvious alternative.

PR #91, opened 2017-04-10
Change some Option() calls to Some() to preserve user nulls

edmundnoble:
Couple of cases where user values might become `None` instead of `null`.


etorreborre:
Oh, how I hate `null`.

PR #92, opened 2017-04-14
Add interpret.trace and some new folds/writer interpreters

edmundnoble:
Just a few Writer utils, as well as interpret.trace for the simplest `Write` implementation.

### DIFF
--- a/shared/src/main/scala/org/atnos/eff/Interpret.scala
+++ b/shared/src/main/scala/org/atnos/eff/Interpret.scala
@@ -212,22 +212,16 @@ trait Interpret {
   /**
    * Interpret the effect T with a side-effect O (see the write method below)
    */
-  def augment[R, T[_], O[_], A](eff: Eff[R, A])(w: Augment[T, O])(implicit m: MemberInOut[T, R]): Eff[Fx.prepend[O, R], A] =  {
-    type U = Fx.prepend[O, R]
-    implicit val mw = MemberIn.MemberInAppendAnyL
-
-    translateInto(eff)(new Translate[T, U] {
-      def apply[X](tx: T[X]): Eff[U, X] = send[O, U, Unit](w(tx)) >> send[T, U, X](tx)
+  def augment[R, T[_], O[_], A](eff: Eff[R, A])(w: Augment[T, O])(implicit memberT: MemberInOut[T, R], memberO: MemberIn[O, R]): Eff[R, A] =  {
+    translateInto(eff)(new Translate[T, R] {
+      def apply[X](tx: T[X]): Eff[R, X] = send[O, R, Unit](w(tx)) >> send[T, R, X](tx)
     })
   }
 
   /**
    * For each effect T add some "log statements" O using the Writer effect
    */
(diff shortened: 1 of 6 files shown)


shared/src/main/scala/org/atnos/eff/WriterEffect.scala
```diff
@@ -94,23 +94,34 @@ trait WriterInterpretation {
...
   }
 
+  def MonoidFold[A : Monoid]: RightFold[A, A] =
+    IntoMonoidFold(identity)
+
   def EvalFold[A](f: A => Eval[Unit]): RightFold[A, Eval[Unit]] = new RightFold[A, Eval[Unit]] {
     type S = Eval[Unit]
     val init = Eval.later(())
+    // should this be Eval.defer rather than later {}.flatten?
+    // are we in danger of accidentally memoizing user side-effects?
```
edmundnoble: @etorreborre ?

shared/src/main/scala/org/atnos/eff/Interpret.scala
```diff
@@ -233,6 +233,13 @@ trait Interpret {
     })
   }
 
+  /**
+   * For a single effect T log every value of that effect
+   */
+  def trace[R, F[_], A](eff: Eff[R, A])(implicit m: MemberInOut[F, R]): Eff[Fx.prepend[Writer[F[_], ?], R], A] =
```
etorreborre: I am not sure about this signature. I think it would be more general to require an implicit `w: Writer[F[_]] |= R`.

shared/src/main/scala/org/atnos/eff/WriterEffect.scala
etorreborre: I have the impression that the 2 expressions should bring the same results.

edmundnoble: What if, for example, the user tries to re-run the Eval they get from EvalFold?

shared/src/main/scala/org/atnos/eff/Interpret.scala
edmundnoble: I was copying from `def writer`. Shall I change the both of them?

etorreborre: I think so, yes.

shared/src/main/scala/org/atnos/eff/WriterEffect.scala
etorreborre: I would need to write a test to check that. Do you think that `later {}.flatten` will memoize?

edmundnoble: I checked in the latest edit; yes, it memoizes. The test I added does not pass with the older version.

shared/src/main/scala/org/atnos/eff/Interpret.scala
edmundnoble: Fixed. In `augment` too.

shared/src/main/scala/org/atnos/eff/WriterEffect.scala
etorreborre: Great, thanks for checking. Can you please remove the 2 lines of comment above then and let's merge.

etorreborre:
I'm merging, I'll fix the comments.

PR #96, opened 2017-05-06
Removed superfluous parameter lists on scalaz either syntax

2chilled:


etorreborre:
What the heck? Thanks for that!

PR #97, opened 2017-05-11
Line up EitherEffectOps implicit members

jarreds:
Noticed that some of the member implicits on EitherEffectOps didn't
match the actual member implicits on the underlying
EitherInterpretation implementation.

This fixes that.


etorreborre:
Thanks.

PR #110, opened 2017-07-17
avoid uncessary fork

taojang:
@etorreborre please review :)


etorreborre:
Good point, I forgot about suspend. I think that should make it.

PR #112, opened 2017-07-18
Added "packing" implicits for member instances

### LINKED ISSUE #56: Provide and document easy to have shorter method signatures
Hi,

I've recently been starting to use eff. And it's amazing, thanks for building this!

However what keeps bugging me is that it leads to fairly long method signatures for monad stacks with many effects

```
def foo[R : _foo : _bar : _baz : ...](myArg: MyInputType): MyOutputType
```

Is there an established best practice for defining aliases for long chains of effect context bounds? If so, could you please add that to the documentation?

Thanks!

etorreborre:
The final solution looks simple, but it took me forever to get there and I'm still quite afraid to test it in more "real" conditions.

@boggle could you please checkout this PR and give it a go on your side?

Review @edmundnoble @benhutchison 

fixes #56

### DIFF
--- a/jvm/src/test/scala/org/atnos/eff/MembersSpec.scala
+++ b/jvm/src/test/scala/org/atnos/eff/MembersSpec.scala
@@ -1,0 +1,75 @@
+package org.atnos.eff
+
+import org.specs2.Specification
+
+class MembersSpec extends Specification { def is = s2"""
+
+ Some signature can be repetitive on some methods using effects (see #56):
+
+   def foo[R :_foo :_bar :_baz](i: Int): Eff[R, Int]
+
+ It is possible to "pack" them with the following type definition:
+
+   type _effects = _foo &: _bar &&: _baz
+
+   def foo[R :_effects](i: Int): Eff[R, Int] =
+     Eff.pure(1)
+
+"""
+
+  def packed = ok
+
+  import Members.{&:, &&:}
+
+  trait Foo[A]
+  trait Bar[A]
+  trait Baz[A]
+  trait Boo[A]
+
+  type _foo[R] = Foo |= R
+  type _bar[R] = Bar /= R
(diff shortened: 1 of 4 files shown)


boggle:
@etorreborre This looks great! I'm leaving for vacation tomorrow but will try to give it a spin soon.

edmundnoble:
This is really, really nice but I actually think it would be more valuable in cats itself. This is a very general-purpose tool.

benhutchison:
👍 Awesome! :) 

More concrete feedback will come as well migrate codebases onto this form.

Edmund, agree it might well find wider usage, but lets get it into Eff now and demonstrate the benefits. Im happy to port to cats version if it gets wider "standardization" later.

dispalt:
whoa, this solves a problem for sure... Ill try to check it out too.

edmundnoble:
Sure @benhutchison, we'll see how it fares here.

etorreborre:
Also my latest version is customized to `eff` and not applicable (for now) to more general contexts. I am indeed guiding the implicit search by indicating that I want to "extract" elements of the form `H Op R` (like `Foo |= R`). If I stay too general and just try to extract `H` I get diverging implicits (again,... How many hours of my life spent on that?).

etorreborre:
@edmundnoble I will wait a bit before merging this for the feedback from other projects. I just want to make sure that:

 - it is working for them
 - it's not introducing a regression for previously working code

etorreborre:
I'm merging this now. Hopefully people will open issues if that doesn't work or create regressions.

boggle:
Thanks for working on this! I gave it a quick spin now and the diverging implicits really get in the way. Will have a second attempt at this but it looks like my current code has too many pieces that require "import Members._" in one line and not in the next.

etorreborre:
Can I say that I'm not really surprised :-)?

PR #115, opened 2017-08-24
Add missing scala.io to the WriterEffect tutorial

richardimaoka:


jvm/src/test/scala/org/atnos/site/lib/WriterEffectPage.scala
```diff
@@ -7,6 +7,7 @@ You can then define your own custom `Fold` to log the values to a file:${snippet
 import org.atnos.eff._, all._, syntax.all._
 import cats.data.Writer
 import java.io.PrintWriter
+import scala.io
```
richardimaoka: http://atnos-org.github.io/eff/org.atnos.site.lib.WriterEffectPage.html

Without this, the following line does not compile.

```scala
io.Source.fromFile("target/log").getLines.toList
```
```
> object Source is not a member of package io
>[error]     val result = io.Source.fromFile("target/log").getLines.toList
>[error]                     ^
```

etorreborre:
Thank you!

PR #116, opened 2017-08-25
DoobieConnectionIO effect

etorreborre:
This is a generalization of @dvic's PR.

In the process of testing this, I think I found a bug for the applicative traversal which amazingly was not showing up until I reworked the `EffSpec`.

If you want to just try it out there is a published version: `4.5.0-20170825100453-3727ee7`.

Please review @dvic and @edmundnoble (and anyone else interested really :-)).


etorreborre:
I'm merging this now because I want to make a new release. Please @dvic come back to me if there's any issue.

dvic:
Sure will do!

PR #118, opened 2017-09-08
Fixed a possible stackoverflow when traversing with Eff applicative

etorreborre:
This fixes some observed stackoverflow on traversals but some concurrency with futures
might be lost. For example (from `EffApplicativeSpec`):
```
Traverse[List].traverse((1 to 500).toList)(i => some[R, Int](i) >> futureDelay(i))
(EffApplicative)
```
However concurrency will be kept with 
```scala
Traverse[List].traverse((1 to 500).toList)(i => some[R, Int](i) *> futureDelay(i))
(EffApplicative)
```

In order to preserve concurrency even with the first example, `Eff.traverseA` has been re-implemented with a restriction. It can only "traverse" containers which are
"indexable" i.e isomorphic to a list of values and indices (see the `Indexable` typeclass).

What do you think @edmundnoble? I tried hard to find better solutions but nothing worked. At first I thought I had a clever solution with representable functors in terms of what can be traversed but those have either fixed or infinite size, hence no lists or vectors.


edmundnoble:
This looks really, really inefficient. It seems to me that what you're doing with `Indexable[F]` is reassociating the `ap` calls; I think the best thing to do here is to rip out the current free applicative formulation and replace it with the new one in 1.0.0-MF. Replacing the ap node's contents with `FreeAp[Union[R, ?], Eff[R, A]]`.

etorreborre:
@edmundnoble I can try that. If I spell out the type completely do you mean 

```
ImpureAp[R, X, Y, A](u: Union[R, X], e: Eff[R, Y], f: (X, Y) => A) extends Eff[R, A]
```
It seems to me that it could be working but I probably need to be careful with the definition of `ap` to avoid stackoverflows.

edmundnoble:
I was more thinking of `ImpureAp[R, X, A](fr: FreeAp[Union[R, ?], Eff[R, A])`, and just using the machinery from `FreeAp` entirely.

etorreborre:
Ok, I see better now, especially because the implementation of `foldMap` is a bit tricky. The things I'm concerned with are:

 - `xs.traverseA(x => OptionEffect.some(x) >> FutureEffect.delay(x))`. Will this keep concurrency for the future effects?
 - I don't see how to support batching

edmundnoble:
From what I can tell yes, that will maintain concurrency.
I'm fairly sure you can do batching via `analyze` or `foldMap`.

etorreborre:
I don't really see how to do batching with foldMap. With batching you have n effects `F` returning n values and you want to transform them in to one call returning a list of values and "re-dispatch" those returned values, to subsequent calls. This transformation actually doesn't interpret the `Eff` action, it just rewrites it, so it is not really a foldMap.

edmundnoble:
So from what I can see you have `n` effects and you need to transform them into `b` effects, where `b` is the batch size. `foldMap` can be used to translate the entire `FreeAp` into a `FreeAp` with less calls.

etorreborre:
I'm closing this attempt for now since I don't have time to work on it

PR #119, opened 2017-09-08
Migrated the code to cats 1.0.0-MF

etorreborre:
and added a new cats-effect module to integrate the IO effect.

For the adventurous this is published as `4.6.1-20170908220345-65de2ec`.

Review @edmundnoble

### DIFF
--- a/monix/shared/src/main/scala/org/atnos/eff/addon/monix/TaskEffect.scala
+++ b/monix/shared/src/main/scala/org/atnos/eff/addon/monix/TaskEffect.scala
@@ -3,7 +3,6 @@ package org.atnos.eff.addon.monix
 import cats._
 import cats.implicits._
 import monix.eval._
-import monix.cats._
 import monix.execution._
 import org.atnos.eff._
 import org.atnos.eff.syntax.all._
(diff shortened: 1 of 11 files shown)


monix/shared/src/main/scala/org/atnos/eff/addon/monix/TaskEffect.scala
```diff
@@ -60,10 +59,12 @@ object TaskCreation extends TaskCreation
 trait TaskInterpretation extends TaskTypes {
 
   private val monixTaskMonad: MonadError[Task, Throwable] =
-    monix.cats.monixToCatsMonadError(Task.typeClassInstances.monadError)
+    MonadError[Task, Throwable]
 
-  private val monixTaskApplicative : Applicative[Task] =
-    monixToCatsApplicative(Task.nondeterminism.applicative)
+  private val monixTaskApplicative : Applicative[Task] = {
+    import Task.nondeterminism
```
edmundnoble: Could this perhaps be reduced to `Task.nondeterminism`?

etorreborre: If I understand correctly `Task.nondeterminism` returns an implicit `ApplicativeStrategy[Task]` which in turn selects the right instance for `Applicative[Task]`.

PR #120, opened 2017-09-26
Retry until a condition becomes true

etorreborre:
This code showed up at code and might benefit others. 

Review @edmundnoble

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,6 @@ cache:
     - $HOME/.sbt/boot/
 
 script:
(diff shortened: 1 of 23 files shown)


edmundnoble:
How does this interact with other effects? Say I have two task-like types in a stack and I retry one, what happens to the other?

etorreborre:
I don't know if anyone is ever mixing different task effects together but I don't think this is going to create any issue. It doesn't really retry one isolated effect but uses the support of any task effect for scheduling delays in order to re-execute the whole computation.

PR #121, opened 2017-09-29
Library upgrades cats 1.0.0

etorreborre:

PR #124, opened 2017-11-03
Transact once per ConnectionIO

jarreds:
The existing `runConnectionIO` method rolls up all `ConnectionIO`
instances to a single DB transaction.

Doobie recommends handling transactions by composing `ConnectionIO`
instances together and that's not possible with the current
`runConnectionIO`.

### DIFF
--- a/doobie/src/main/scala/org/atnos/eff/addon/doobie/DoobieConnectionIOEffect.scala
+++ b/doobie/src/main/scala/org/atnos/eff/addon/doobie/DoobieConnectionIOEffect.scala
@@ -5,7 +5,7 @@ import java.sql.Connection
 import cats.implicits._
 import cats.{MonadError, ~>}
 import _root_.doobie.free.connection.ConnectionIO
-import _root_.doobie.imports.Transactor
+import _root_.doobie.imports._
 import org.atnos.eff._
 import org.atnos.eff.all._
 
@@ -21,7 +21,7 @@ trait DoobieConnectionIOCreation extends DoobieConnectionIOTypes {
 
 trait DoobieConnectionIOInterpretation extends DoobieConnectionIOTypes {
 
-  def runConnectionIO[R, U, F[_], E, A, B](e: Eff[R, A])(t: Transactor[F])(
+  def runConnectionIOCombine[R, U, F[_], E, A, B](e: Eff[R, A])(t: Transactor[F])(
     implicit mc: Member.Aux[ConnectionIO, R, U],
              mf: F /= U,
              me: MonadError[F, E] ): Eff[U, A] = {
@@ -57,6 +57,17 @@ trait DoobieConnectionIOInterpretation extends DoobieConnectionIOTypes {
 
       interceptErrors(before >> runEffect(connection) << after)(oops).addLast(send(always))
     }
+  }
+
+  def runConnectionIO[R, U, F[_], E, A, B](e: Eff[R, A])(t: Transactor[F])(
+    implicit mc: Member.Aux[ConnectionIO, R, U],
+             mf: F /= U,
+             me: MonadError[F, E]): Eff[U, A] = {
+
+    interpret.translate(e)(new Translate[ConnectionIO, U] {
+      def apply[X](c: ConnectionIO[X]): Eff[U, X] =
+        send(c.transact(t))
+    })
   }
 }
 
--- a/doobie/src/main/scala/org/atnos/eff/syntax/addon/doobie.scala
+++ b/doobie/src/main/scala/org/atnos/eff/syntax/addon/doobie.scala
@@ -15,6 +15,13 @@ trait doobie {
 }
 
 final class DoobieConnectionIOOps[R, A](val e: Eff[R, A]) extends AnyVal {
+  def runConnectionIOCombine[F[_], U, E, B](t: Transactor[F])(
+    implicit mc: Member.Aux[ConnectionIO, R, U],
+    mf: MemberInOut[F, U],
+    me: MonadError[F, E]): Eff[U, A] = {
+    DoobieConnectionIOInterpretation.runConnectionIOCombine[R, U, F, E, A, B](e)(t)
+  }
+
   def runConnectionIO[F[_], U, E, B](t: Transactor[F])(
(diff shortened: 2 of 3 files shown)


jarreds:
I have some tests based on the branch in https://github.com/atnos-org/eff/pull/121 that I will add once that's merged.

doobie/src/main/scala/org/atnos/eff/addon/doobie/DoobieConnectionIOEffect.scala
```diff
@@ -58,6 +58,44 @@ trait DoobieConnectionIOInterpretation extends DoobieConnectionIOTypes {
...
+    def getConnection: Eff[U, Connection] =
+      send[F, U, Connection](t.connect(t.kernel))
+
+    def interceptErrors[Y](effect: Eff[U, Y])(oops: F[Unit]): Eff[U, Y] =
+      interpret.interceptNat(effect)(new (F ~> F) {
+        def apply[X](f: F[X]): F[X] =
+          f.handleErrorWith((err: E) => oops *> me.raiseError[X](err))
+      })
+
+    interpret.translate(e)(new Translate[ConnectionIO, U] {
```
etorreborre: Is there a way to factor out that code that's similar to the above method into one place?

jarreds: Definitely. Will update shortly.

edmundnoble:
I'm not following the rationale here. 
"Doobie recommends handling transactions by composing ConnectionIO
instances together and that's not possible with the current
runConnectionIO."

"The existing runConnectionIO method rolls up all ConnectionIO
instances to a single DB transaction."

This seems to be contradictory. What is the difference between "composing ConnectionIO instances together" and "rolling up all ConnectionIO instances to a single DB transaction"?

If you don't have all of the `ConnectionIO` in a single transaction, isn't that going to a) break transactional semantics and b) cause a performance problem because of all the connections?

etorreborre:
I think we eventually want the flexibility to interpret the `ConnectionIO` effect in different ways:

 1. start a transaction at the beginning of the interpretation and use it to interpret all `ConnectionIO` values inside the same transaction. This is the behaviour of the current interpreter. This allows to compose independent functions returning more than just a `ConnectionIO` effect (using a logging effect for example) and still get just one transaction

 2. start one transaction for each `ConnectionIO` effect to interpret. This is the addition contained in the current proposal

 3. let the user decide transaction boundaries but providing `startTransaction` and `endTransaction` primitives. This could be implemented in a separate PR

jarreds:
Thanks @edmundnoble. Yeah, I need to clarify.

> What is the difference between "composing ConnectionIO instances together" and "rolling up all ConnectionIO instances to a single DB transaction"?

Composing `ConnectionIO` instances would be where we can control which queries are part of a db transaction/conn:

```scala
//  runConnectionIO
val foo: Eff[R, A] = for {
  a <- fromConnectionIO(query1.flatMap(query2)) // one transaction/conn for these two queries
  b <- fromIO(longRunningBlockingTask(a)) // takes a few seconds
  c <- fromConnectionIO(query3(b).flatMap(query4)) // a diff transaction/conn for these two
} yield c
```

In this case, two separate transactions/conn are established with the DB. One for each `ConnectionIO`. Neither connections sat idle waiting for `b`.

With the existing functionality, all `ConnectionIO` instances are rolled up to one DB transaction/conn -- this is what I've renamed `runConnectionIOCombine`:

```scala
// runConnectionIOCombine
val foo: Eff[R, A] = for {
  a <- fromConnectionIO(query1.flatMap(query2))  // database transaction/conn opened here
  b <- fromIO(longRunningBlockingTask(a)) // takes a few seconds
  c <- fromConnectionIO(query3(b).flatMap(query4)) // runs in the same transaction/conn as `a`
} yield c
```

In this case, the DB transaction/conn sits open idle while `b` is running and all four queries are committed together.

I'd like to have the option that could satisfy either of the scenarios. `runConnectionIO` for the former and `runConnectionIOCombine` for the latter.

jarreds:
> 3. let the user decide transaction boundaries but providing startTransaction and endTransaction primitives. This could be implemented in a separate PR

That would be really nice.

etorreborre:
Note that the cool thing about the Eff approach is that you are not blocked right now! You can always provide your own interpreter, this is a nice benefit.

edmundnoble:
I think 3. is probably the most useful thing here; already I think the approach in 2. smells, and I've come to the conclusion that why I think this is because one of the obvious and intuitive (to me) laws that Eff normally follows no longer holds:
```scala
lift(fa) >>= (f.andThen(lift)) == lift(fa >>= f) (up to observation with `detach`)
```

Which is also part of the definition of a free monad.

jarreds:
@edmund, yeah, agree on that smell.

This is from the doobie docs:

> __How do I do several things in the same transaction?__
> You can use a for comprehension to compose any number of ConnectionIO programs, and then call .transact(xa) on the result. All of the composed programs will run in the same transaction. For this reason it’s useful for your APIs to expose values in ConnectionIO, so higher-level code can place transaction boundaries as needed.

We have an existing query library that is based on that model, and I currently can't use that library with the existing  Eff `ConnectionIO` effect because it ignores the transaction boundaries.

A couple questions for ya:

* Regarding option 2, can you think of any way we could integrate the concept from doobie of a connection/transaction boundary per ConnectionIO program without breaking that law you referenced? Or is adding new transaction boundaries in Eff our only option?
* Any ideas on a lawful way we could pull off option 3?

doobie/src/main/scala/org/atnos/eff/addon/doobie/DoobieConnectionIOEffect.scala
jarreds: Was able to simplify this a lot and remove the duplicated code.

BenHutchisonSeek:
@jarreds Just checking in... Are you still using the Eff-Doobie integration? Is this PR still useful to you?

jarreds:
Nope, not using this particular integration anymore. Feel free to close if you'd like.

PR #126, opened 2018-01-12
Fix #93: Move site docs from core to eff project 

### LINKED ISSUE #93: Explain how to set Monix ExecutionModel of a Task containing an Eff-subprogram
Current Eff API allows putting an Eff sub-program into a Monix Task effect:
`def taskSuspend[R :_task, A](task: =>Task[Eff[R, A]], timeout: Option[FiniteDuration] = None): Eff[R, A]`

But I am unclear how to obtain precise control of "forking" - specifically [ExecutionModel](https://monix.io/docs/2x/execution/scheduler.html#execution-model) - for Eff tasks. Motivation is to allow tasks collections to execute concurrently.

- Discussion on [gitter](https://gitter.im/atnos-org/eff?at=58f1cef6408f90be66c12aa7) showed how to fork *all* Tasks in an Eff program. [Experiment confirmed concurrent execution works](http://scastie.org/30537).

- But Monix docs [describe forking as inserting an "Async boundary"](https://monix.io/docs/2x/eval/task.html#taskfork--taskasyncboundary) between two tasks. An [attempt to write this as an Eff program](http://scastie.org/30535) did not execute concurrently. The idea is:
```scala
fromTask(Task.unit.asyncBoundary) >> taskDelay(println(s"Im on ${Thread.currentThread}"))
```

Would like to understand & document how to write an Eff program with an Async Boundary that enables concurrent execution of Eff-subprograms.

It not clear that this is an Eff "framework", versus "Eff ecosystem" issue, but Im entering it here to keep track of this open question.

benhutchison:
So that the classpaths of extensions are visible. Add some docs for Monix task.

Notes: I'd previous assumed that a `docs` sub-project made sense, but on examination of the sbt build structure, I realized that the `eff` parent project already has the visibility of both core and the extensions that the docs require. So if the site specs site under `eff` project (ie live at `src/test/scala`) it works.

This requires
-  a classpath dependency on the test code in `coreJVM` from the site specs in `eff`
- including JVM settings in eff project, since its now doing compilation.


etorreborre:
That's great, thanks @benhutchison !

PR #135, opened 2018-04-22
Upgrade monix to 3.0.0-RC1, cats effect to 0.10.1.

benhutchison:
There were quite a few deprecated APIs as both Monix and Cats effect are rapidly evolving. Most migrations were exactly as per the deprecation advice.

One choice Im unsure about was [casting `Applicative[Task.Par]` to `Applicative[Task]`](https://github.com/atnos-org/eff/compare/master...benhutchison:monixupdate?expand=1#diff-e2389f0077c98d9bb90fdb1ab4893dc6R72). It seems like a clash of techniques: Eff's Task addon had chosen to explicitly pass the instances, while Monix had chosen to use a newtype wrapper around Task.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,13 +9,13 @@ import sbtrelease._
 import org.scalajs.jsenv.nodejs._
 
 lazy val catsVersion        = "1.0.1"
-lazy val monixVersion       = "3.0.0-M3"
+lazy val monixVersion       = "3.0.0-RC1"
 lazy val scalazVersion      = "7.2.7"
 lazy val specs2Version      = "4.0.2"
 lazy val twitterUtilVersion = "17.11.0"
 lazy val catbirdVersion     = "0.21.0"
 lazy val doobieVersion      = "0.5.0"
-lazy val catsEffectVersion  = "0.9"
+lazy val catsEffectVersion  = "0.10.1"
 lazy val fs2Version         = "0.10.2"
 
 lazy val eff = project.in(file("."))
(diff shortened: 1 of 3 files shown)


etorreborre:
I am surprised that the casting works because `Task != Task.Par`. I would vote for writing an `Applicative[Task]` instance locally:
```scala
private val applicativeTask extends Applicative[Task] {
  override def ap[A, B](ff: Task[(A) => B])(fa: Task[A]): Task[B] =
    Task.mapBoth(ff, fa)(_ (_))

  override def map2[A, B, Z](fa: Task[A], fb: Task[B])(f: (A, B) => Z): Task[Z] =
    Task.mapBoth(fa, fb)(f)

  override def product[A, B](fa: Task[A], fb: Task[B]): Task[(A, B)] =
    Task.mapBoth(fa, fb)((_, _))

  override def pure[A](a: A): Task[A] =
    Task.now(a)

  override val unit: Task[Unit] =
    Task.now(())

  override def map[A, B](fa: Task[A])(f: (A) => B): Task[B] =
    fa.map(f) 
}
```

etorreborre:
Hi @benhutchison, btw I want to thank you for all your efforts maintaining eff! I just gave you admin rights on the repo if you want / need more freedom to move things around.

benhutchison:
Eric Ive pushed the local instance. 

However I do think my cast worked OK, here's why:`Task.Par` is just a [newtype wrapper](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/Task.scala#L3190) around Task, which selects this [parallel applicative](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/instances/CatsParallelForTask.scala#L48) similar to the local instance, ie uses `mapBoth`.

etorreborre:
Oh, I didn't get that this was an unboxing newtype. I still prefer your update to be on the safe side.

etorreborre:
👍

PR #140, opened 2018-04-27
modifyS 

benhutchison:
`modifyS.using` is a State `modify` variant that often enables better type inference for the modify function `f`(when used in an effect stack with a single state effect).

It does this by capturing the type params `R` and `S` before receiving the `f: S => S` parameter.

Discussion: There's potentially a huge number of similar ergonomic improvements to the Eff API. It's debatable whether they should go into Eff core or an extension library. I can go either way. @etorreborre WDYT?

I do find that the amount of type annotation needed to make Eff work is cumbersome. Hence Im investing effort into trying to find idioms that work well with scalas type inference (like this example). Probably more to come..

### DIFF
--- a/jvm/src/test/scala/org/atnos/eff/StateEffectSpec.scala
+++ b/jvm/src/test/scala/org/atnos/eff/StateEffectSpec.scala
@@ -21,6 +21,8 @@ class StateEffectSpec extends Specification with ScalaCheck { def is = s2"""
 
  The Eff monad is stack safe with State $stacksafeState
 
+ modifyS is a modify variant that enables better type inference for state modification closures $modifySState
+
 """
 
   def putGetState = {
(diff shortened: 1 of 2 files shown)


etorreborre:
I think that if it is a consistent set of idioms (a bit like there used to be `traverse` and `traverseU` in Scalaz) it is fine to add that to core and document it. It is a shame that we have to do that because it make the API surface larger but I think it's better than cumbersome type annotations.

etorreborre:
👍 but the User Guide should probably be extended with a note on type inference for those effects.

benhutchison:
On consideration: 

+1 to the goal of a "consistent set of idioms".

However, I dont think we've found those idioms yet. I think we need to see more examples and more experiments. At this stage, I want the freedom to try stuff, change my mind etc.

Therefore, I suggest keeping extensions like `modifyS` out of core, where source compatibility is a strong constraint, until they're proven.

Instead, create an incubator project, with a charter that it serves as a proving ground for less stsbale features that might make it to core. I think that model has proven itself in open source.

I have a little project [Effx](https://github.com/benhutchison/effx) Im thinking to use as an incubator. So far its been personal sharing only, but I'll add a README and a link to it.

benhutchison:
Closing, may revisit an experimental module again but I don't have enough material to warrant it ATM

PR #148, opened 2018-08-14
Publish cats effect module for Scalajs. Upgrade cats-effect

benhutchison:
- Publish cats effect module for Scalajs. 
- Upgrade cats-effect dependency to 1.0.0-RC2

Also @etorreborre, while I was doing this I noticed this [strange dependency from Scalaz -> Cats Effect](https://github.com/atnos-org/eff/blob/master/build.sbt#L72) - is it a mistake, or intended?


benhutchison:
! not ready for merge. Discovered that the source code in catsEffect module isn't getting compiled because it isn't under `shared/`

etorreborre:
No this dependency looks like a mistake to me.

benhutchison:
OK, the upgrade to cats effect `1.0.0-RC2` is problematic. Specifically implementing the instances of `Async[Eff[R, ?]]` and `Effect[Eff[R, ?]]`, as both typeclasses have grown in scope and complexity.

I got `asyncF` to compile, but at the heavy price of needing to lift the dubious `runIO: Eff[R, Unit] => IO[Unit]` up to Async as well. But there's also `bracketCase` and `runSyncStep` unsatisfied.

@etorreborre Would you consider calling these typeclass instances a "failed experiment" and dropping them? 

I know that might sound a big step at first glance, but I suspect many users of Eff, including myself, actually have no need of the instances. All I want is to include `IO` in my stack, but Im happy to interpret them, in the same way eg Monix Task or Futures are. I don't need Eff to *be* an IO-like thing.

And I dont think the typeclass contracts can be fulfilled by Eff anyway, already `runIO` is a bit of a lie.

etorreborre:
Yes I think this clarifies the situation and also helps us answer the question about using Eff as the base monad for fs2, the answer is 'no'.

PR #155, opened 2018-12-03
Ior in validate effect

buzden:
`Validate` effect was changed to be able to be interpreted as `cats.Ior`.

Special warning construction functions were added for constructing errors that are not fatal (equivalent to `Ior`'s `Both`). Treatment of such warnings in existing `Either`/`Validated` interpreters has been made to be the same as what `Ior.toEither` does.


etorreborre:
I haven't used warnings with Validate before but that seems like a nice use case 💯

PR #157, opened 2018-12-04
Some effects creation functions were changed to have by-name arguments

buzden:
In creation functions of some of effects arguments were made to be by-name in case when some of arguments may be unneeded.


etorreborre:
Good points.

PR #159, opened 2018-12-11
Added memoize for cats-effect

vitold:
Hi, I added memoize to the IOEffects, we plan to use it during our huge refactoring, so it would be awesome if you review the changes and if it's fine - release a new version. 
Thanks!

### DIFF
--- a/cats/shared/src/main/scala/org/atnos/eff/addon/cats/effect/IOEffect.scala
+++ b/cats/shared/src/main/scala/org/atnos/eff/addon/cats/effect/IOEffect.scala
@@ -86,7 +86,7 @@ trait IOInterpretation extends IOTypes {
     */
   def ioMemo[R, A](key: AnyRef, cache: Cache, e: Eff[R, A])(implicit task: IO /= R): Eff[R, A] =
     ioAttempt(Eff.memoizeEffect(e, cache, key)).flatMap {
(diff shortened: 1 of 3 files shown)


cats/shared/src/main/scala/org/atnos/eff/addon/cats/effect/IOEffect.scala
```diff
@@ -72,6 +74,58 @@ trait IOInterpretation extends IOTypes {
...
+    interpret.translate(effect)(new Translate[Memoized, U] {
+      def apply[X](mx: Memoized[X]): Eff[U, X] =
+        mx match {
+          case Store(key, value) => IOEffect.ioDelay(cache.memo(key, value()))
+          case GetCache() => IOEffect.ioDelay(cache)
+        }
+    })
+  }
+
+  implicit val taskSequenceCached: SequenceCached[IO] = new SequenceCached[IO] {
```
etorreborre: This should be called `ioSequenceCached` isn't it?

etorreborre:
Sorry for the late answer, please do the small suggestion and I will publish today.

cats/shared/src/main/scala/org/atnos/eff/addon/cats/effect/IOEffect.scala
vitold: yep :)

etorreborre:
Thanks!

etorreborre:
I released `5.4.1`

PR #161, opened 2018-12-17
Simpler implementation of applicative effect for errors catching

buzden:
I suggest a simpler implementation of applicative effect interpreter of the `ValidateEffect` 's `catchWrong` interpreter.

The idea of simplification came by observing the existing implementation and realizing that `catchWrong` actually catches only the first one. Thus, internal state on applicative can be represented as a simple `Either` actually.

---

What I suggest also (but afraid to do due to source incompatibility) is **to rename** the current `catchWrong` to `catchFirstWrong` because it loses all errors except the first one like in cases where

- we have several consequent `wrong` calls or
- when we have some traversable effect like `List` interpreted before the `Validate` effect, i.e. like `runList.catchWrong(whatever).runNel.run`).

Yet, it is the simplest known catcher thus, maybe, still useful.

---

Universal catcher should have a choice to give the right value, to ignore this error in order to let the catcher to work with other errors or to combine the result of catching of an error with other errors catcher's results. While the last seems to be doable (still having problems or whether being able to recatch previously thrown error inside the catcher or ont), the second looks yet a little bit too error-prone to implement.

Still, I want to discuss whether or not existing `catchWrong` can be renamed or not.

### DIFF
--- a/jvm/src/test/scala/org/atnos/eff/ValidateEffectSpec.scala
+++ b/jvm/src/test/scala/org/atnos/eff/ValidateEffectSpec.scala
@@ -94,7 +94,7 @@ class ValidateEffectSpec extends Specification with ScalaCheck { def is = s2"""
         a <- EffMonad[S].pure(3)
       } yield a
 
-    validate.catchWrong((s: String) => pure(4)).runNel.run ==== Right(4)
+    validate.catchFirstWrong((s: String) => pure(4)).runNel.run ==== Right(4)
   }
 
   def catchWrongValues2 = {
@@ -105,8 +105,8 @@ class ValidateEffectSpec extends Specification with ScalaCheck { def is = s2"""
     val handle: E => Check[Unit] = { case e => tell[Comput, E](e).as(()) }
 
     val comp1: Check[Int] = for {
-      _ <- wrong[Comput, E]("1").catchWrong(handle)
-      _ <- wrong[Comput, E]("2").catchWrong(handle)
+      _ <- wrong[Comput, E]("1").catchFirstWrong(handle)
+      _ <- wrong[Comput, E]("2").catchFirstWrong(handle)
     } yield 0
 
     val comp2: Check[Int] = comp1
--- a/jvm/src/test/scala/org/atnos/eff/ValidateEffectSpec.scala
+++ b/jvm/src/test/scala/org/atnos/eff/ValidateEffectSpec.scala
@@ -19,19 +19,23 @@ class ValidateEffectSpec extends Specification with ScalaCheck { def is = s2"""
  recover from wrong values                   $catchWrongValues1
  recover from wrong values and tell errors   $catchWrongValues2
 
- recover, check that the first is catched    $catchFirstWrongValue
- recover, check that all are catched, simple $catchAllWrongValuesLinear
- recover, check that the last is catched     $catchLastWrongValue
- recover, the whole list is catched          $catctListOfWrongValues
+ recover from several, monadic
+   the first is catched                      ${ForCatchingEffMonadic.catchFirstWrongValue}
+   all are catched                           ${ForCatchingEffMonadic.catchAllWrongValues}
+   the last is catched                       ${ForCatchingEffMonadic.catchLastWrongValue}
 
- recover, check the first catched, applic    $catchFirstWrongValueApplicative
- recover, check all are catched, applicative $catchAllWrongValuesApplicative
- recover, check the last catched, applic     $catchLastWrongValueApplicative
+ recover from several, applicative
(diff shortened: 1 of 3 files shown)


buzden:
> the current implementation catches the _last_ `Wrong` value (I think)

The first time I was looking at the implementation, I thought the same. However after playing with it, I realized that it catches the first, after which I thought that `State[Option, ?]`'s `Applicative` instance follows the behaviour from `Option` which is fail-fast (thus, catches the first).

> This suggests an implementation when you pass a `Semigroup` which is either be `First[A]`, `Last[A]`, `List[A]` and have shortcut methods as `catchFirstWrong`, `catchLastWrong`, `catchAllWrongs`. Then we alias `catchWrong` to `catchFirstWrong` and deprecate it because it is ambiguous. Does that answer your question?

The idea seems to be great! One thing seems to be somewhat strange: `Semigroup[A]` gives `(A, A) => A` and it is the right case for `First[A]` and `Last[A]`. For `List[A]` we want a way for pack `A` to `List[A]`, i.e. kind of `pure`. Since we don't have a typeclass for pointed set, it seems we need to get `Applicative` also. So, the final signature should be somewhat

```
def catchWrongs[R, E, A, T[_]: Applicative](effect: Eff[R, A])(handle: T[E] => Eff[R, A])(implicit member: Validate[E, ?] <= R, semi: Semigroup[T[E]])
```

where for `catchFirstWrong` `T` would be `Id`, thus `T[E]`'s `Semigroup` would be `E`'s Semigroup, but at the same time, we can have `T` to be `List` for `catchAllWrongs` (and have a list concatenation as a `Semigroup`).

That's a pity that we don't have multiple implicit arguments lists in Scala 2 (so, the call of such method with the explicitly given semigroup would look like `catchWrongs(effect)(handler)(implicitly, First[E])` or whatever like this). But if is this thought seems to be reasonable, I can try to implement such generalized catcher.

etorreborre:
Thanks, I will have a look at your PR tomorrow!

jvm/src/test/scala/org/atnos/eff/ValidateEffectSpec.scala
```diff
@@ -105,15 +109,63 @@ class ValidateEffectSpec extends Specification with ScalaCheck { def is = s2"""
...
+        _ <- ValidateEffect.correct[S, String, Int](1)
+        _ <- ValidateEffect.wrong[S, String]("error1")
+        a <- EffMonad[S].pure(3)
+        _ <- ValidateEffect.wrong[S, String]("error2")
+      } yield a.toString
+
+    validate.catchAllWrongs((ss: NonEmptyList[String]) => pure(ss.mkString_("", ", ", ""))).runNel.run ==== Right("error1, error2")
+  }
+
+  def catchAllWrongValuesApplicative = {
```
etorreborre: The name `Applicative` here might be a bit misleading. I was thinking this was about the `Applicative` instance of `Eff` not the `Applicative` instance of the container used to collect the values. Which makes me think that we should add some tests for `Eff` `Applicative` cases:
```
val action = ValidateEffect.wrong[S, String]("error1") |@| ValidateEffect.wrong[S, String]("error1")
// what does that return?
validate.catchAllWrongs . ...
```

buzden: I agree, so, I've renamed this test, added three more and reorganized all a little bit to make everything more clean and to reduce duplications.

buzden: Oh, and by the way, new applicative tests introduced one more warning which I do not have an idea how to fix (and as far as I know, it is a compiler bug which is fixed only in dotty). So, sorry.

shared/src/main/scala/org/atnos/eff/ValidateEffect.scala
```diff
@@ -121,38 +121,65 @@ trait ValidateInterpretation extends ValidateCreation {
...
-
     })
+
+  /** catch and handle the first wrong value */
+  def catchFirstWrong[R, E, A](effect: Eff[R, A])(handle: E => Eff[R, A])(implicit member: Validate[E, ?] <= R): Eff[R, A] = {
+    implicit val first: Semigroup[E] = Semigroup.instance{ (a, _) => a }
+    catchWrongs[R, E, A, Id](effect)(handle)
+  }
+
+  /** catch and handle the first wrong value */
```
etorreborre: s/first/last

buzden: Done

etorreborre:
Cool, thank you let's merge!

PR #164, opened 2019-02-16
Support Monix 3.0.0-RC2

dispalt:


etorreborre:
Thanks, I can see there is still a compilation error but this does not seem too hard to fix.

dispalt:
Fixed @etorreborre

etorreborre:
Thank you!

PR #169, opened 2019-04-28
Adopt the Scala Code of Conduct

rossabaker:
Typelevel is about to announce that it is moving to the Scala Code of Conduct for all participating projects. The Scala Code of Conduct was by developed by the Scala Center with input from Typelevel and improves on the Typelevel code of conduct in a number of ways and can be thought of as the "Typelevel code of conduct v. 2.0". We want people participating in projects and activities right across the Scala ecosystem to be able to expect a uniform standard of good behavior, and coordinating on a shared code of conduct is an important step in that direction.


etorreborre:
Thanks @rossabaker  for the PR

PR #172, opened 2019-05-05
Scala 2.13.0-RC2

xuwei-k:


scalaz/src/test/scala/org/atnos/eff/addon/scalaz/concurrent/TaskEffectSpec.scala
```diff
@@ -65,7 +65,7 @@ class TaskEffectSpec(implicit ee: ExecutionEnv) extends Specification with Scala
     } yield a + b
 
     eventually(retries = 5, sleep = 0.seconds) {
-      action[S].taskAttempt.runOption.runSequential must returnValue(beSome(beLeft(boomException)))
+      action[S].taskAttempt.runOption.runSequential must returnValue(Some(Left(boomException)))
```
xuwei-k: workaround ambiguous implicits. (Maybe specs2 issue? 🤔 )

```
[error]  found   : => Throwable
[error]  required: org.specs2.matcher.Expectable[?]
[error] Note that implicit conversions are not applicable because they are ambiguous:
[error]  both method theValue in trait MustExpectations1 of type [T](t: => T)org.specs2.matcher.MustExpectable[T]
[error]  and method thisValue in trait ShouldExpectations of type [T](t: => T)org.specs2.matcher.ShouldExpectable[T]
[error]  are possible conversion functions from => Throwable to org.specs2.matcher.Expectable[?]
[error]       action[S].taskAttempt.runOption.runSequential must returnValue(beSome(beLeft(boomException)))
[error]                                                                                    ^
[error] one error found
[error] (Test / compileIncremental) Compilation failed
```

shared/src/main/scala/org/atnos/eff/Member.scala
```diff
@@ -42,8 +42,8 @@ trait MemberInLower2 extends MemberInLower3 {
   implicit def MemberIn3L[L[_], M[_], R[_]]: MemberIn[L, Fx3[L, M, R]] =
     TaggedMemberIn(1)
 
-  implicit def MemberInAppendL[T[_], L, R](implicit append: MemberIn[T, L]): MemberIn[T, FxAppend[L, R]] =
-    AppendMemberIn(isRight = false, append)
+  implicit def MemberInAppendAnyL[T[_], R]: MemberIn[T, FxAppend[Fx1[T], R]] =
+    AppendMemberIn(isRight = false, TaggedMemberIn(1))
```
xuwei-k: swap implicit def due to ambiguous 😕

scalaz/src/test/scala/org/atnos/eff/addon/scalaz/concurrent/TaskEffectSpec.scala
etorreborre: This is annoying. Something has definitely changed in the implicit resolution. `beLeft` has 2 forms: one taking a parameters and another one directly returning a matcher. In 2.13 Scala seems to assume that we want the second form and tries to use the `.apply` method. I did a fix for this in specs2 but in the meantime in `eff` a workaround is to annotate a type and write `beSome(beLeft[Throwable](boomException))`.

shared/src/main/scala/org/atnos/eff/Member.scala
etorreborre: This is one place where I am very scared with implicits and `eff`. The whole towering of implicits is very ad-hoc.

xuwei-k: fixed

etorreborre:
Thank you, if I have a bit of time I'll do the 2.13-RC3 update

PR #178, opened 2019-07-08
Update dependencies

xuwei-k:


symbiont-eric-torreborre:
Thanks, I find it weird that Monix still hasn't a final 3.0.0 version, it feels like it has been on milestones and RCs for a long time now

PR #184, opened 2019-08-04
Add support for ZIO.

takayahilton:


BenHutchisonSeek:
On gitter there's been a conversation around the future maintenance of Eff, with @etorreborre wanting to step back as IIUC he's working in Haskell these days.

So, I want to suggest this is better published as a standalone library than a part of the Eff repo.. I dont use ZIO myself and are reluctant to add another extension module now - if anything I'd rather split them out to reduce future maintenance.

Thats not a reflection of the PR quality itself, as the code looks quite correct.

I'd be happy to link to it in the docs/README

PR #189, opened 2019-09-10
Add up to 20 fx types

carlosdagos:
Hello,

This PR adds more `fx` `type`s, up to 20. 

Reason being that past 12 effects, a current solution is to use `append`/`prepend`. I think this will make that look cleaner.

This clearly isn't the most elegant way, so I'm happy to close this if there's more favorable way of doing it.

I don't know why the limit was set to 12 before. If that was an actual limitation in the past then maybe this won't work for all consumers, but I tested and compiled this locally and seemed to work fine.

Thank you!


etorreborre:
The limitation to 12 is totally arbitrary. However it should also make you consider if are properly layering abstractions into your application. Maybe some "low-level" effects should not "bubble-up" to the top. You should rather have a few top-level effects interpreted in terms of low-level ones.

PR #224, opened 2020-01-13
upgrade to cats 2.1

SethTisue:
there is some minor-looking compile-time error here

can I interest you/somebody in finishing this PR for me? it would
enable me to keep eff in the Scala community build (where we are
moving to cats 2.1)

(eff is a leaf project in the community build at present (no downstream
projects depend on it), so it's a would-be-nice, not a blocker)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import org.scalajs.jsenv.nodejs._
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
 
-lazy val catsVersion        = "2.0.0"
+lazy val catsVersion        = "2.1.0"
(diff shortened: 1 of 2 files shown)


benhutchison:
Looks like the problem is that Eff still supports Scala 2.11 while cats 2.1.0 does not. I'm not sure if anyone still needs 2.11 support so maybe it can be dropped

SethTisue:
fwiw, I'm increasingly seeing Scala OSS projects dropping 2.11, that's become pretty common (a minority, perhaps, but a now quite sizable minority)

benhutchison:
Ive upgraded Eff to use cats 2.1 in release 5.7.0

PR #292, opened 2020-10-06
Replace applicative methods for composing last actions to `flatMap`

y-yu:
- Since #290, any last action is not gone if we call `addLast` several times
- It uses applicative methods (`map2`, `*>` and `<*`) to compose last action `Eff`s, the order of execution for last actions is up the interpreter
- atnos-eff's `Interperter` interface allows us to implement the special applicative case as `onApplicativeEffect`
    - so we can improve our interpreter's efficiency
- However, I don't agree that last actions (which added by `addLast`) should be executed applicatively, I want to know the order of the execution from code
- For example, in the current implementation `addLast` to the `Eff[Task, A]` is not the same of that of `Eff[Eval, A]`
- Because I change applicative methods to `flatMap`, and then `Eff[Task, A]` and `Eff[Eval, A]` are the same
    - I think it's simpler than before 🤔 
- Discussions are very welcome!


y-yu:
I'll make a new function: `addLastM` which is monadic version of the current `addLast` rather than modifying. So I close this PR.

PR #381, opened 2021-05-19
[scala3 compat] Enable implicit-conversion explicitly

exoego:
This will resolve warnings on Scala 3

 ```
[warn] -- Feature Warning: /Users/vagrant/IdeaProjects/eff/shared/src/main/scala/org/atnos/eff/syntax/error.scala:11:21 
[warn] 11 |  implicit final def toErrorOrOkOps[A](c: Error Either A): ErrorOrOkOps[A] = new ErrorOrOkOps(c)
[warn]    |                     ^
[warn]    |Definition of implicit conversion method toErrorOrOkOps should be enabled
[warn]    |by adding the import clause 'import scala.language.implicitConversions'
[warn]    |or by setting the compiler option -language:implicitConversions.
```

### DIFF
--- a/shared/src/main/scala/org/atnos/eff/syntax/choose.scala
+++ b/shared/src/main/scala/org/atnos/eff/syntax/choose.scala
@@ -1,5 +1,7 @@
 package org.atnos.eff.syntax
 
+import scala.language.implicitConversions
+
 import cats.Alternative
 import org.atnos.eff._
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: shared/src/main/scala/org/atnos/eff/syntax/choose.scala [hunk 1]
```diff
@@ -1,13 +1,15 @@
...
+
 import cats.Alternative
 import org.atnos.eff._
 
 object choose extends choose
 
 trait choose {
   implicit final def toChooseEffectOps[R, A](e: Eff[R, A]): ChooseEffectOps[R, A] = new ChooseEffectOps(e)
-}
+}˚
```
xuwei-k, line 12: 👀

PR #391, opened 2021-06-06
Fix maybe a typo? AuthErroEither in the document

shomatan:
Hello. Thanks for the awesome library.
I found a typo(?) in the [official documentation](https://atnos-org.github.io/eff/org.atnos.site.TransformStack.html#translate-an-effect-into-multiple-others)
Please point out if there are any mistakes 🙇 
Thank you.

### DIFF
--- a/org.atnos.site.TransformStack.html
+++ b/org.atnos.site.TransformStack.html
@@ -409,8 +409,8 @@ sealed trait Authenticated[A]
 case class Authenticate(token: String) extends Authenticated[AccessRights]
 type _authenticate[U] = Authenticated |= U
 
-type AuthErroEither[A] = AuthError Either A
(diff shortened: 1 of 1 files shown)


shomatan:
Sorry, I'm probably wrong. Should I fix this one?
https://github.com/atnos-org/eff/blob/master/src/test/scala/org/atnos/site/TransformStack.scala#L153-L154

benhutchison:
Hi @shomatan , that symbol has an odd name, but as as you can see, it is used on the next line to define `_error` and not used again. So the impact is small. Renaming it as you suggest makes sense, but if so should be done in all occurrences.

shomatan:
Thanks for the reply.
I would like to close this PR once as it is the wrong way to fix it:smile:

